Issue #15, opened 2022-12-08
[AIP-4][Discussion] Update Simple Map To Save Gas


areshand:
# AIP: Update Simple Map To Save Gas

## Summary

Change the internal implementation of SimpleMap in order to reduce gas prices with minimal impact on Move and Public APIs.

## Motivation

The current implementation of SimpleMap uses a BCS-based logarithmic comparator identifying slots on where to store data in a Vector. Unfortunately this is substantially more expensive than a trivial linear implementation, because each comparison requires BCS serialization followed by a comparison. The conversion to BCS cannot be resolved as quickly as traditional comparator can and substantially impacts gas prices.

## Proposal

- Replace the internal definition of `find` from the current logarithmic implementation to a linear search across the vector.
- Replace the functionality within `add` to call `vector::push_back` and append new values instead of inserting them into their sorted position.

## Rationale

This will result in the following performance differences:

| Operation | Gas Unit Before | Gas Unit After | Delta |
| --- | --- | --- | --- |
| CreateCollection | 174200 | 174200 |  |
| CreateTokenFirstTime | 384800 | 384800 |  |
| MintToken | 117100 | 117100 |  |
| MutateToken | 249200 | 249200 |  |
| MutateTokenAdd10NewProperties | 1148700 | 390700 | 64% |
| MutateTokenMutate10ExistingProperties | 1698300 | 411200 | 75% |
| MutateTokenAdd90NewProperties | 20791800 | 10031700 | 51% |
| MutateTokenMutate100ExistingProperties | 27184500 | 10215200 | 62% |
| MutateTokenAdd300NewProperties (100 existing, 300 new) | 126269000 | 135417900 | -7% |
| MutateTokenMutate400ExistingProperties | 143254200 | 136036800 | 5% |

When the token only has 1 property on-chain, we can see the mutation token cost doesn’t change. However, 

if the user wants to add 10 new properties or update existing properties, the gas cost is reduced by **64% and 75%.** 

if the user wants to store 100 properties on-chain, the gas cost is reduced by 51% and 62%.

600 property mutations were also tested, but failed due to exceeding maximum gas.

## Implementation

Draft benchmark PR https://github.com/aptos-labs/aptos-core/pull/5765/files

```rust
// Before the proposed change
public fun add<Key: store, Value: store>(
        map: &mut SimpleMap<Key, Value>,
        key: Key,
        value: Value,
  ) {
      let (maybe_idx, maybe_placement) = find(map, &key);
[...]

PR #16, opened 2022-12-08
[AIP-5] Soulbound and Freezing Token Standard

JacobADevore:


aips/aip-4.md
```diff
@@ -0,0 +1,78 @@
...
+- Monitoring a single event store for soulbound and non-soulbound tokens
+- Compatibility with the Aptos-Token standard.
+
+An alternative solution to the non-transferability problem is creating a "Locking Standard", where tokens can be deposited, claimed, and burned. Creating a locker that acts as a wrapper around tokens essentially makes them soulbound. This solution has a significant drawback: monitoring can no longer be handled for both soulbound and non-soulbound tokens in a single event store, thus creating the need for separate monitoring of an entirely different event store. This is undesirable from usability, composability, and scalability standpoints.
+
+### Specification
+
+Modifications required to the current [Aptos-Token](https://github.com/aptos-labs/aptos-core/blob/main/aptos-move/framework/aptos-token/sources/token.move) implementation:
+
+- Append “soulbound: bool” too *Tokens* and *Collections.*
```
areshand, line 29: what is the usage of soulbound collections?

aips/aip-4.md
```diff
@@ -0,0 +1,78 @@
...
+- Minting of soulbound tokens can only be handled through *“claims”*. After a soulbound NFT has been claimed it is locked to the receiver with the only valid operation being *“burn”,* if applicable to the token’s configuration*.*
+- Validate before any transfer execution to ensure the token isn’t soulbound.
+
+```rust
+/// Executed when a specific token_id is set or updated to soulbound, according to the parameter "bounded".
```
areshand, line 34: does this mean a token can mutate between a soulbound and non-soulbound token? In my mind, soulbound should not be transferrable after minting to its owner

aips/aip-4.md
```diff
@@ -0,0 +1,78 @@
+# AIP-4: Multi-Composable Token Standard
```
areshand, line 1: Thank you for submitting the PR. please also create a discussion issue like this one https://github.com/aptos-foundation/AIPs/issues/15. It is easier to keep the discussion in an issue

aips/aip-4.md
JacobADevore: After further thought, I do not believe Collections requires the soulbound property. Good call.

aips/aip-4.md
JacobADevore: No, the soulbound property should never be mutatable. The validation checks the existing transfer functions to ensure that the current token is not soulbound. This validation is needed as soulbound and non-soulbound tokens can coexist.

aips/aip-4.md
```diff
@@ -0,0 +1,78 @@
...
+
+An alternative solution to the non-transferability problem is creating a "Locking Standard", where tokens can be deposited, claimed, and burned. Creating a locker that acts as a wrapper around tokens essentially makes them soulbound. This solution has a significant drawback: monitoring can no longer be handled for both soulbound and non-soulbound tokens in a single event store, thus creating the need for separate monitoring of an entirely different event store. This is undesirable from usability, composability, and scalability standpoints.
+
+### Specification
+
+Modifications required to the current [Aptos-Token](https://github.com/aptos-labs/aptos-core/blob/main/aptos-move/framework/aptos-token/sources/token.move) implementation:
+
+- Append “soulbound: bool” too *Tokens*
+- Minting of soulbound tokens can only be handled through *“claims”*. After a soulbound NFT has been claimed it is locked to the receiver with the only valid operation being *“burn”,* if applicable to the token’s configuration*.*
+- Validate before any transfer execution to ensure the token isn’t soulbound.
```
davidiw, line 31: simply verify the withdraw

aips/aip-4.md
```diff
@@ -0,0 +1,78 @@
...
+	}
+
+}
+```
+
+**** This function allows multiple offers to be sent in a single transaction.
+
+### Suggested I**mplementation Timeline**
+
+To be determined.
```
davidiw, line 67: I think this is one of those things where you should try a couple implementations and come back with the pros and cons. It isn't clear to me that there is a perfect solution here.

davidiw (commented):
Really exciting to have our first proposal from the community. I look forward to working with you more on this topic :)

aips/aip-4.md
JacobADevore: Will do, working through it with @areshand :)

aips/aip-4.md
JacobADevore: > Really exciting to have our first proposal from the community. I look forward to working with you more on this topic :)

Thank you @davidiw for all the amazing feedback here!!

aips/aip-4.md
sherry-x: Hey @JacobADevore could you please add the header like other AIPs? Let's make this one AIP-5. once you updated, we can land this draft and iterate on it

JacobADevore: > Hey @JacobADevore could you please add the header like other AIPs? Let's make this one AIP-5. once you updated, we can land this draft and iterate on it

Sure thing! Does that suffice?

sherry-x: You need to update the file name too

JacobADevore: Got it! Sorry about that.

aips/aip-4.md
subashmokshya: > what is the usage of soulbound collections?

1. Gift cards
2. Birthday Wishes
3. Lifetime community memberships

These are a few that i can think at this moment but there're more.

e99243506bigplay:
`0x3::token` is really gas heavy enough, why we should keep change that?
I recommend an another approach: clone `0x3::token` standard to be `0x3::soulbound_token`, so you don't need do a verify for withdraw/deposit.

alnoki:
@JacobADevore  I suggest renaming "Soulbound" to either "SignerBound" or "KeyBound" for the following reasons:

1. Distances a useful blockchain primitive from quasi-religious nomenclature, which could prove exclusionary for individuals with (a)religious/philosophical attitudes that reject unfalsifiable metaphysical claims about souls. Notably the notion of a soul has historically been abused, most frequently in the service of denying that certain classes of sentient beings lack them (slaves, women, heathens, non-human animals) such that these classes can be subjected to harmful treatment without reproach. Consider that GitHub has changed the default branch name from `master` to `main` to avoid references to similarly violent historical institutions.
2. Clarifies that the token is indeed bound to a key/signer, not the individual person behind the signature, which will (a) avoid confusion in the instance that an account is compromised and (b) serves to increase a security-minded mindset around not losing a private key (as opposed to assuming that someone can "recover" their SingerBound token through appeal to a central Web2 authority, for instance).
3. Avoids confusion that may arise with having a token affiliated with a multisignature wallet. Here, the term "KeyBound" may be more appropriate than "SignerBound", since there are multiple signers on a multisig wallet, though notably the account is itself treated as having a single conglomerate signature. Either option, however, provides more clarity than the initial name in this proposal, which raises the question of whose soul the token belongs to from the collection of signatories (or perhaps there is an aggregate soul drifting somewhere above the entire group?).
4. Provides more appropriate description within the context of Aptos' key rotation schema, which can theoretically support the transfer of such a token from one person to another: here, someone just has to rotate their authentication key to someone else's public key. And actually, now that I think about it, I suspect that Aptos' key rotation schema might actually prohibit the kind of bound token originally put forth here. If I have some special token in my account, what is to stop someone from paying me paper fiat to rotate the authentication key for the account to a public key that they specify?

Understandably such a renaming may be met with some resistance given that the term "soulbound" has already achieved some amount of prominence, but it is still early enough to alter the trajectory of this norm into a more inclusive and contextually-meaningful term. To this end, I'd especially like to hear your thoughts on reason 1. Also, is it correct that the discussion in reason 4 raises a potentially-insuperable technical challenge?

davidiw:
With the progress towards token v2 that supports this inherently and with minimal cost, do we still want to pursue this?

e99243506bigplay:
Why should be so hard? just do `move_to(&signer, YourTokenName {})` then make it able to query from another contract, next to frozen the contract then, the SBT is made out.

davidiw:
> Why should be so hard? just do `move_to(&signer, YourTokenName {})` then make it able to query from another contract, next to frozen the contract then, the SBT is made out.
> 
> Of course, by this method, the user need to interact the contract, but this contract has been frozen and open source, 100% safe,

@e99243506bigplay, I agree, the goal of this AIP was to make this inline and a part of the existing core code, which would have ramifications to every token. If someone wants to push a TokenV1 AIP that supports soul bound as you proposed, I think that would be acceptable from a framework perspective -- the question is would it be good for @JacobADevore and would other ecosystem partners adopt.

davidiw:
> > > Why should be so hard? just do `move_to(&signer, YourTokenName {})` then make it able to query from another contract, next to frozen the contract then, the SBT is made out.
> > > Of course, by this method, the user need to interact the contract, but this contract has been frozen and open source, 100% safe,
> > 
> > 
> > @e99243506bigplay, I agree, the goal of this AIP was to make this inline and a part of the existing core code, which would have ramifications to every token. If someone wants to push a TokenV1 AIP that supports soul bound as you proposed, I think that would be acceptable from a framework perspective -- the question is would it be good for @JacobADevore and would other ecosystem partners adopt.
> 
> Hey @davidiw, I think it would be great if you can extend or fork your https://github.com/aptos-labs/aptos-core/tree/main/aptos-move/move-examples/token_objects project and show an example of how soul bound NFTs would come to life with the new object pattern.

Take a look here:
https://github.com/aptos-labs/aptos-core/blob/main/aptos-move/framework/aptos-token-objects/sources/aptos_token.move#L189

Any objections to closing this and focusing on TokenV2? There are already many wallets that support it: Petra, Fewcha, Rise, and more coming soon :)

thepomeranian:
Closing this for now due to inactivity and no actionable items

PR #29, opened 2023-01-07
[AIP] token framework reserved properties

areshand:


aips/aip-8.md
```diff
@@ -0,0 +1,57 @@
...
+## Summary
+
+This proposal introduces framework-reserved properties to
+
+1. prevent properties used by the token framework being changed by creators freely.
+2. reserving properties that can be extended to support programmable token behavior ex: token freezing, soulbound token, etc
+
+## Motivation
+
+We have existing token properties used by our token standard to control who can burn the token. However, when the token’s default properties are mutable, creators can add these control properties after the token has been minted. The creator can burn these tokens from collectors. This is a known issue called out in token standard to set the token default properties to be immutable as a best practice.  To prevent this entirely, this proposal is to make it infeasible to update the control properties after the token creation.
```
davidiw, line 21:
```suggestion
We have existing token properties used by our token standard to control who can burn the token. However, when the token’s default properties are mutable, creators can add these control properties after the token has been minted. The creator can burn these tokens from collectors. This is a known issue called out in the token standard to set the token default properties to be immutable as a best practice.  To prevent this entirely, this proposal is to make it infeasible to update the control properties after the token creation.
```
please add link to the token standard

aips/aip-8.md
```diff
@@ -0,0 +1,57 @@
...
+// this function will be called by token mutation methods to add or update token properties
+public(friend) fun add_non_framework_reserved_properties(map: &mut PropertyMap, key: String, value: PropertyValue)
+public(friend) fun update_non_framework_reserved_properties(map: &mut PropertyMap, key: String, value: PropertyValue)
+```
+
+## Risks and Drawbacks
+
+**Overhead of token default property mutation cost**
+
+The validation of framework reserved properties will cost extra gas when calling `mutate_tokendata/token_property` , Currently, when creators mutate the properties, the function loop through all the properties to be mutated and update the property value.  The additional cost would be we need to check if the property key starts with `TOKEN_`.  This cost is very minimal. The current function already did many validations of the string length, key duplication, etc, the overhead of checking if a key has a prefix should have negligible impact on user experience.
```
davidiw, line 49:
```suggestion
The validation of framework reserved properties will cost extra gas when calling `mutate_tokendata/token_property` , Currently, when creators mutate the properties, the function loop through all the properties to be mutated and update the property value.  The additional cost would be we need to check if the property key starts with `TOKEN_`.  This cost should be minimal by using substring matching and early exit of comparisons. Also the current function already does many validations on the string length, key duplication, etc, the overhead of checking if a key has a prefix should have negligible impact on gas costs and user experience.
```

aips/aip-8.md
```diff
@@ -0,0 +1,57 @@
...
+
+## Timelines
+
+The change is to be implemented. Ideally, these should be landed to mainnet in 1/2 weeks and waiting for mainnet release.
```
davidiw, line 53:
```suggestion
The change has yet to be implemented. Ideally, these should will landed to main in 1 to 2 weeks, then testnet, followed by mainnet.
```

aips/aip-8.md
```diff
@@ -0,0 +1,57 @@
...
+
+## Future Potentials
+
+There will be follow-up AIPs on soulbound token and token freezing leveraging these framework-reserved properties.
```
davidiw, line 57:
```suggestion
This unblocks follow-up AIPs on soulbound token and token freezing leveraging this framework-reserved properties.
```

davidiw (approved):
let's not number an AIP before it has been approved

also not you called this aip-8.md but list it as aip 9 here

let's fix up and land

aips/aip-8.md
```diff
@@ -0,0 +1,57 @@
...
+## Summary
+
+This proposal introduces framework-reserved properties to
+
+1. prevent properties used by the token framework being changed by creators freely.
+2. reserving properties that can be extended to support programmable token behavior ex: token freezing, soulbound token, etc
+
+## Motivation
+
+We have existing token properties used by our token standard to control who can burn the token. However, when the token’s default properties are mutable, creators can add these control properties after the token has been minted. The creator can burn these tokens from collectors. This is a known issue called out in the token standard to set the token default properties to be immutable as a best practice.  To prevent this entirely, this proposal is to make it infeasible to update the control properties after the token creation.
```
davidiw, line 21: we need a references section or something to point to the token spec

davidiw:
Thought I left a comment on the number.. let's not put a number up there until it is ready for community review
The numbers are also off in that the doc says 8 but the content says 9.

aips/aip-8.md
areshand: added

areshand:
> Thought I left a comment on the number.. let's not put a number up there until it is ready for community review The numbers are also off in that the doc says 8 but the content says 9.

updated to "TBD". I think the aip-7 has the number 8, which was the reason I used 9.

sherry-x (approved):

PR #35, opened 2023-01-15
Update resource_groups.md

davidiw:
Adding some updates after wrapping up the first complete implementation:
* Took in Alden's suggestion on renaming the attribute fields
* Cleaned up some wonky text that had somehow gotten in
* Tightened up some of the writing for better readability

AIP seems like it is probably ready for community review. Will discuss with gatekeepers.

### DIFF
--- a/aips/resource_groups.md
+++ b/aips/resource_groups.md
@@ -31,6 +31,10 @@ A resource group co-locates data into a single storage slot by encoding within t
 
 At the storage layer, the resource groups are stored as a BCS-encoded BTreeMap where the key is a BCS-encoded fully qualified struct name (`address::module_name::struct_name`, e.g., `0x1::account::Account`) and the value is the BCS-encoded data associated with the resource.
 
+![image](https://user-images.githubusercontent.com/73818/212690642-f8c24ed8-8869-4ce2-9941-4958aae3f8a9.png)
+
+The above diagram illustrates data stored at address `0xcafef00d`. `0x1::account::Account` is a resource stored at address `0xcafef00d`. `0xaa::resource::Group` contains a set of resources or a resource group stored at the same address. The resource group packs multiple resources into the group. Resources within a resource group require nested reading, wherein first the resource group must be read from storge followed by reading the specific resource from the resource group.
+
 ### Alternative 1 — Any within a SimpleMap
 
 One alternative that was considered is storing data in a `SimpleMap` using the `any` module. While this is a model that could be shipped without any change to Aptos-core, it incurs some drawbacks around developer and application complexity both on and off-chain. There’s no implicit caching, and therefore any read or write would require a deserialization of the object and any write would require a serialization. This means a transaction with 3 writes would result in 3 deserializations and 3 serializations. In order to get around this, the framework would need substantial, non-negligible changes, though with the emergence of `SmartMap` there may be more viability here. Finally, due to the lack of a common pattern, indexers and APIs would not be able to easily access this data.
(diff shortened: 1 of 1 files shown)


aips/resource_groups.md
```diff
@@ -99,6 +99,8 @@ During compilation and publishing, these attributes are checked to ensure that:
 
 From a storage perspective, a resource group is stored as a BCS-encoded `BTreeMap<StructTag, BCS encoded MoveValue>`, where a `StructTag` is a known structure in Move of the form: `{ account: Address, module_name: String, struct_name: String }`.  Whereas, a typical resource is stored as a `BCS encoded MoveValue`.
 
+Resource groups introduce a new storage access path: `ResourceGroup` to distinguish from existing access paths. This provides a cleaner interface and segration of different types of storage. This becomes advantageous to indexers and other direct readers of storage that can now parse storage without inspecting module metadata.
```
movekevin, line 102: Might be a bit lengthy but can we also clarify that since the current APIs only allow accessing specific resources via Resource access path or modules via Module access path, there's no way to access a specific resource group and load all of the contained resources. These individual resources, however, can be accessed via the list account resources API or by referring to each of these resources specifically. There are no use cases for fetching resource groups and they can be viewed as an internal data structure for smart contracts and VM.

movekevin (approved):

aips/resource_groups.md
```diff
@@ -75,31 +73,33 @@ The motivation of using a `struct` is that
...
 ```
 
 During compilation and publishing, these attributes are checked to ensure that:
 
-1. A `resource_group_container` has no abilities and no fields.
-2. The `scope` within the `resource_group_container` can only become more permissive, that is it can either remain at a remain at the same level of accessibility or increase to the next.
-3. Each entry within a resource group has a `resource_group` attribute.
-4. The `container` parameter is set to a struct that is labeled as a `resource_group_container`.
-5. During upgrade, an existing `struct` cannot either enter or leave a `resource_group`.
+1. A `resource_group_member` has no abilities and no fields.
```
banool, line 92: This is a typo right? The members must have fields to be useful? (as they do above).

aips/resource_groups.md
```diff
@@ -75,31 +73,33 @@ The motivation of using a `struct` is that
 2. It can build upon the existing storage model that knows how to read and write data stored at  `StructTag`s. Thus it limits the implementation impact to the VM and readers of storage, storage can remain agnostic to this change.
 3. Only `struct` and `fun` can have attributes, which in turn let’s us define additional parameters like `scope`.
 
-Each entry in a resource group is identified by the `resource_group` attribute:
+Each entry in a resource group is identified by the `resource_group_member` attribute:
 
 ```move
-#[resource_group(container = aptos_framework::object::ObjectGroup)]
+#[resource_group_member(group = aptos_framework::object::ObjectGroup)]
 struct Object has key {
```
banool, line 80: If this `resource_group_member` attribute is present, does it mean we'll enforce that these structs can only be used as part of that resource group, and not in any other resource group / on its own?

aips/resource_groups.md
```diff
@@ -75,31 +73,33 @@ The motivation of using a `struct` is that
...
-1. A `resource_group_container` has no abilities and no fields.
-2. The `scope` within the `resource_group_container` can only become more permissive, that is it can either remain at a remain at the same level of accessibility or increase to the next.
-3. Each entry within a resource group has a `resource_group` attribute.
-4. The `container` parameter is set to a struct that is labeled as a `resource_group_container`.
-5. During upgrade, an existing `struct` cannot either enter or leave a `resource_group`.
+1. A `resource_group_member` has no abilities and no fields.
+2. The `scope` within the `resource_group_member` can only become more permissive, that is it can either remain at a remain at the same level of accessibility or increase to the next.
+3. Each entry within a resource group has a `resource_group_member` attribute.
+4. The `group` parameter is set to a struct that is labeled as a `resource_group`.
+5. During upgrade, an existing `struct` cannot either enter or leave a `resource_group_member`.
```
banool, line 96: I read the AIP but I'm still not sure why we have these constraints? I figure it's so we don't have to deal with backfilling / null data respectively?

banool (commented):
In the risks and drawbacks you mention the API layer. I figure you mean mostly the API <-> storage interface, rather than the data that the API serves? If not, it might be good to have more info there about what you expect the changes to look like.

aips/resource_groups.md
davidiw: yes, will make sure that is clear in the writing.

aips/resource_groups.md
davidiw: ugh, yes, this is resource_group

aips/resource_groups.md
davidiw: you'll end up in a situation where data is lost, I'll make that clearer. Specifically, if a struct enters a resource group any existing data will become inaccessible. Now one could theoretically add key to something after the fact and potentially make that into a resource group. But key + resource group go hand in hand.

aips/resource_groups.md
davidiw: sure

Issue #33, opened 2023-01-09
[AIP-9][Discussion] Higher-Order Inline Functions for Collections


wrwg:
aip | title | author | discussions-to | Status | last-call-end-date | type | created | updated
-- | -- | -- | -- | -- | -- | -- | -- | --
9 | Higher-Order Inline Functions for Collections | wrwg | #33 | Draft | TBD | Standard (Framework) | 2023/1/9 | 2023/1/9

[...]

PR #36, opened 2023-01-15
[AIP] Higher-Order Inline Functions for Collections

wrwg:
In this AIP, we suggest a set of conventions to add higher-order inline functions to collection types as they appear in the Move and Aptos frameworks.

Discussion issue #33

### DIFF
--- a/aips/higher-order-functions.md
+++ b/aips/higher-order-functions.md
@@ -66,7 +66,7 @@ public inline fun map<T, S>(v: vector<T>, f: |T|S): vector<S> {
 
 ### Effected Data Types in Aptos
 
-Those data types in the Aptos frameworks should get the higher-order functions:
+Those data types in the Aptos frameworks should get the higher-order functions (TO BE COMPLETED):
 
 - Move stdlib
     - vector
@@ -97,8 +97,9 @@ None visible.
 
 ## Future Potential
 
-- The Move Prover can benefit from these functions to avoid loop invariants which are one of the most challenging parts of working with the prover.
-- In the future, function parameters may also be supported by the Move VM. Then this proposal simply generalizes such that it does not longer depend on inline functions.
+- Parts of the Aptos framework can be rewritten to make them better auditable by removing low-level loops and replacing them by calls to higher-order functions.
+- The Move Prover can benefit from these functions to avoid looplow-level which are one of the most challenging parts of working with the prover.
+- In the future, function parameters may also be supported by the Move VM. Then this proposal simply generalizes such that higher-order functions on collections not necessarily need to be inline functions.
 
 ## Suggested implementation timeline
 
--- a/aips/higher-order-functions.md
+++ b/aips/higher-order-functions.md
@@ -1,0 +1,105 @@
+---
+aip: TBD
+title: Higher-Order Inline Functions for Collections
+author: wrwg
+discussions-to: https://github.com/aptos-foundation/AIPs/issues/33
+Status: Draft
+last-call-end-date: TBD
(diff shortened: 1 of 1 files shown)


e99243506bigplay:
ah no, incoming more gas fee on operation.

aips/higher-order-functions.md
```diff
@@ -0,0 +1,132 @@
...
+type: Standard (Framework)
+created: 2023/1/9
+updated: 2023/1/9
+---
+
+## Summary
+
+Recently, the concept of *inline functions* has been added to Aptos Move.
+Those functions are expanded at compile time and do not have an equivalent
+in the Move bytecode. This allows them to implement a feature which is
```
davidiw, line 17: "this" and "them" don't read clearly
might recommend something like
"This offers the ability to implement"
or
"Inline functions allow us to implement a feature currently unavailable for ..."

aips/higher-order-functions.md
```diff
@@ -0,0 +1,132 @@
...
+currently not available for regular Move functions: taking functions,
+given as lambda expressions, as parameters. Given this, we can
+define popular higher-order functions like for_each, filter, map, and fold
+collection types in Move. In this AIP, we suggest a set of conventions
+to add those functions to the data structure in the Move and Aptos frameworks.
```
davidiw, line 22: "to the data structure" sounds more foundational -- like all data structures
I think
"to existing data structures in"

aips/higher-order-functions.md
```diff
@@ -0,0 +1,132 @@
...
+
+- Only collection types which are expected to be iterable in a single
+  transaction should offer these functions. For example, tables are not such
+  collections.
+- Some collection types may diverge by adding less or more of those
+  functions, depending on the type of the collection.
+
+### Effected Data Types in Aptos
+
+Those data types in the Aptos frameworks should get the higher-order functions:
```
davidiw, line 98: it might be worth calling out that the expectation that in the future it is an expectation that other data structures will implement them as a prerequisite for inclusion

davidiw (approved):
I'd recommend considering to place an entire paragraph on a single line rather than cutting this up like code:
* Git is actually pretty good about giving deltas
* In this format, each time someone makes an update, it results in massive shifts across the text
* Everyone seems to specify a different line length amount

This is how we tend to do it in aptos.dev, and we can also discuss it in the AIP sync to maybe align on a standard.

aips/higher-order-functions.md
```diff
@@ -0,0 +1,105 @@
...
+
+## Reference Implementation
+
+TODO: link to the code in the framework at `main` once the PRs landed
+
+## Risks and Drawbacks
+
+None visible.
+
+## Future Potential
```
movekevin, line 98: We should also consider migrating use cases within the Aptos Framework to these new inline functions as well. For example: https://github.com/aptos-labs/aptos-core/blob/35976bfa00bc06d376285da38739ee86f8a96d75/aptos-move/framework/aptos-framework/sources/stake.move#L997

aips/higher-order-functions.md
```diff
@@ -0,0 +1,105 @@
...
+
+```move=
+public inline fun map<T, S>(v: vector<T>, f: |T|S): vector<S> {
+    let result = vector<S>[];
+    for_each(v, |elem| push_back(&mut result, f(elem)));
+    result
+}
+```
+
+### Effected Data Types in Aptos
```
movekevin, line 67: nit: Effected -> Affected

movekevin (approved):

PR #37, opened 2023-01-15
Update objects.md

davidiw:
* removed aggregable types -- let's just make that another spec
* added WithdrawRef that empowers the creator to withdraw as appropriate, this enables assisted use of tokens
* added DepositRef that can be generated up to a certain point in time, this enables assisted use of tokens
* added some API specs


movekevin (approved):

aips/objects.md
```diff
@@ -22,10 +22,10 @@ The object model allows Move on Aptos to represent a complex type as a set of re
 The object model improves type safety and data access through the use of a capability framework implemented by different accessors or references that define and distribute out various capabilities that allow for rich operations. These include:
 
 - `CreatorRef` that allows the creation of all other capabilities, allows adding resources to the object, and can only be accessed at the time the object is created
-- `OwnerRef` defines ownership of an object. Objects may have one or more `OwnerRef` and an `OwnerRef` can be aggregated, thus allow for fungability of objects.
-- `MintRef` allows for the creation of `OwnerRef`.
+- `OwnerRef` defines ownership of an object.
 - `DeleteRef` allows the holder to delete the object from storage.
```
CapCap, line 26: can we add more granularity here for removing from an object group (or is that party of `ExtendRef`?) vs deleting the object as a whole? What if someone puts a table into the object group?

aips/objects.md
```diff
@@ -135,15 +152,34 @@ struct WithdrawEvent has drop, store {
...
-public fun withdraw(account: &signer, addr: address): OwnerAbility;
+public fun withdraw(account: &signer, addr: address): OwnerRef;
 
-public fun deposit(account: &signer, object: OwnerAbility);
-```
+public fun withdraw_with_ref(withdraw_ref: &WithdrawRef): OwnerRef;
+
+public fun deposit(account: &signer, object: OwnerRef);
+
+public fun deposit_with_ref(deposit_proof: &DepositProof, object: OwnerRef);
```
CapCap, line 161: what is this and why?

aips/objects.md
davidiw: That would require ExtendRef. We could write more, but I think that might be better for giving a BaseObject example?

aips/objects.md
```diff
@@ -43,22 +42,23 @@ The existing Aptos data model emphasizes the use of the `store` ability within M
...
+- Simplified storage interface for heterogeneous types.
 - Enable a rich and dynamic data model while limited impact on storage, i.e., storage should not take up more than effectively a single resource (storage slot): see Resource Groups AIP.
 - Data is immobile but ownership still remains portable: storing data in resources and ownership within `OwnerRef`.
 - Make easily usable from entry functions: each object is stored at an address, which is a supported API type.
 - Offer rich set of capabilities: the `Ref` model.
 - Support emitting events directly: directly built-in.
 - Simplified access to storage avoiding costly deserialization, serialization: resources, once loaded, have limited costs associated with future access.
-- Objects are fully deletable.
-- Objects must be fully instantiated at creation time and cannot have new resources added afterward
+- Objects are deletable.
```
CapCap, line 60: is this setting a flag, or dropping everything inside the group and setting a flag?

aips/objects.md
```diff
@@ -67,20 +67,21 @@ Objects are built with the following considerations:
 - This is repeated until the top-most ancestor is the `Object` struct, which defines the `create_object` function.
 - The `create_object` generates a new address, stores an `Object` struct at that address within the `ObjectGroup` resource group, and returns back a `CreatorRef`.
 - The create functions called earlier can use the `CreatorRef` to obtain a signer, store an appropriate struct within the `ObjectGroup` resource, make any additional modifications, and return the `CreatorRef` back up the stack.
-- In the object creation stack, any of the modules can define properties such as ownership, deletion, and mutability.
-- Upon finishing the object creation, the `OwnerRef` can be stored in an `ObjectStore`, where it can be then transferred to another `ObjectStore` or even embedded within another object, offering object composability.
+- In the object creation stack, any of the modules can define properties such as ownership, deletion, transferability, and mutability.
 
 ### The Core Object
 
 An object is stored in the `ObjectGroup` resource group. The core `Object` contains a single field: `guid_creation_num`, which is used to create distinct event streams.
```
CapCap, line 74: no longer accurate

aips/objects.md
```diff
@@ -90,60 +91,50 @@ Each object is stored in its own address or object id. To ensure unique Ids, the
...
-    - Data layout: `{ self: address, amount: u64 }`
-- `MintRef`
-    - Has the ability to generate `OwnerRef`
-    - Abilities: `store`
-    - Data layout: `{ self: address }`
+    - Data layout `{ self: ObjectId }`
 - `DeleteRef`
     - Can be used to remove an object from the `ObjectGroup`
     - There can be many for a single object
     - They cannot be aggregated
```
CapCap, line 106: Why many/not aggregatable? Are any other aggregatable? Do we not expect other ref types to have multiple for an object?

aips/objects.md
```diff
@@ -90,60 +91,50 @@ Each object is stored in its own address or object id. To ensure unique Ids, the
...
+### API
 
 ```move
-struct ObjectStore has key {
-    inner: table::Table<address, OwnerRef>,
-    deposits: event::EventHandle<DepositEvent>,
-    withdraws: event::EventHandle<WithdrawEvent>,
-}
+public fun create_object(creator: &signer, seed: vector<u8>): CreatorRef;
+public fun generate_object_id(object_id: address): OwnerRef;
```
CapCap, line 122: is `OwnerRef` still meaningful?

aips/objects.md
```diff
@@ -13,22 +13,21 @@ requires: AIP_number_of(Resource Groups)
...
 
 The object model improves type safety and data access through the use of a capability framework implemented by different accessors or references that define and distribute out various capabilities that allow for rich operations. These include:
 
-- `CreatorRef` that allows the creation of all other capabilities, allows adding resources to the object, and can only be accessed at the time the object is created
-- `OwnerRef` defines ownership of an object. Objects may have one or more `OwnerRef` and an `OwnerRef` can be aggregated, thus allow for fungability of objects.
-- `MintRef` allows for the creation of `OwnerRef`.
+- `CreatorRef` that allows the creation of all other capabilities, allows adding resources to the object, and can only be accessed at the time the object is created.
```
CapCap, line 24: the other refs can be stored, and in sum do the same thing basically that `CreatorRef` do, except without being able to get the signer. Can we please consider adding `store` to `CreatorRef` as well? I understand there is a design pattern you are thinking of here, but to give creators options, having `store` might be the most flexible

aips/objects.md
```diff
@@ -13,22 +13,21 @@ requires: AIP_number_of(Resource Groups)
...
+- `ObjectId` points to an object. This is useful for storing references to a resource for reverse lookup and other operations.
 - `DeleteRef` allows the holder to delete the object from storage.
-- `ExtendRef` allows for the holder to add new resources to the object after creation.
+- `TransferRef` allows for the holder to transfer new resources to the object after creation.
```
CapCap, line 27: Is this for moving structs into the object?

Is there a ref that allows one to change the objects owner, then?

aips/objects.md
```diff
@@ -152,15 +143,14 @@ There is an early reference implementation in the following PR: https://github.c
 
 Open areas for discussion include:
 
-- Whether or not to offer the ability for objects to have new resources added after creation? The current proposal has `ExtendRef` that can call `move_to`
-- Should `Object` be single owner and should that single owner be placed in the `Object` itself. As a result do we need `amount` in `OwnerRef` or `OwnerRef` at all. Note, we’d likely want some form of read-only Ref, so that if an object owns another object, it can be determined by reading on-chain only data.
 - Should addition and deletion of resources count the number of resources within an object to ensure that deletion is complete?
-- Should ownership be aggregable?
-- Should we block this on the availability of `BucketTable` or something similar for tracking ownership
+- How explicit should this AIP be about any of the abilities? Should there be a BaseObject for trivial usage that dictates an owner-centric set of policies.
+- Deletion could be complete, but then we need to ensure two objects can never be stored at the same address. An objected created over another risks corrupting storage due to overlapping event streams.
```
CapCap, line 148: for sure; what if deletion did:
1. delete the object (and everything in it)
2. create a new empty object in the same spot
3. set `deleted` to `true`, which would make it immutable

aips/objects.md
davidiw: Considering that there's really only two ref types that creator can leverage (Delete and Transfer), I think this isn't worth spending too much time on at this juncture.

A major reason to not allow creator ref is that it provides the same backdoor functionality of allowing a creator of an nft to specify that an nft can be burned after mint.

I think it is important to make it clear intentionality when creating something. If I claim something is immutable and forever the data and the API associated with that object should be easily verifiable as such.

aips/objects.md
davidiw: I'm not sure what this means "removing from an object group" -- using an object group is optional -- I should clarify that above.

aips/objects.md
davidiw: This is the ref that allows the holder to arbitrarily move the object, mind you for most objects, this will be unnecessary and they can use the owner transfer paradigm.

aips/objects.md
davidiw: I should just delete this...

aips/objects.md
```diff
@@ -67,83 +68,115 @@ Objects are built with the following considerations:
...
+    - Optionally used to provide a link of ownership from the owner to the object
+    - Abilities: `store`
+    - Data layout: `{ self: ObjectId }`
+- `TransferRef`
+    - Can be used transfer ownership of the object from one address to another
+    - Abilities: `drop, store`
+    - Data layout `{ self: ObjectId }`
+- `LinearTransferRef`
+    - Used to actually perform the transfer
+    - Enforces an entity can only transfer once, assuming that they do not have direct access to `TransferRef`
```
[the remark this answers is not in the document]
davidiw: This enables gating at a higher level

aips/objects.md
davidiw: No storage refund... We could offer two models

PR #42, opened 2023-01-20
Rename aip-0.md to aip-0-AIP Process.md

michelle-aptos:


sherry-x:
Hey there, have space in file name is very anti-pattern, let's not do that. I think keeping the AIP title table in the repo readme should be sufficient. Ideally we would have a page on foundation website to show the status and have prettier view.

davidiw (approved):

PR #51, opened 2023-01-24
[aip-11] Minimal edits

wrwg:
Removing a statement about `any` which isn't true (or misunderstandable?), as long as `any` is used in a linear timeline of compatible upgrades in a given network. Otherwise looks great!

### DIFF
--- a/aips/aip-11-Move Objects.md
+++ b/aips/aip-11-Move Objects.md
@@ -7,13 +7,14 @@ Status: Draft
 last-call-end-date:
 type: Standard (Framework)
 created: 2023/01/05
-updated: N/A
-requires: AIP_number_of(Resource Groups)
+updated: 2023/01/23
+requires: [AIP-9](./aip-9-Resource Groups.md) (Resource Groups)
 ---
 
 ## Summary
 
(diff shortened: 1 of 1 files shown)


aips/aip-11-Move Objects.md
```diff
@@ -36,7 +37,7 @@ The existing Aptos data model emphasizes the use of the `store` ability within M
 
 - Data cannot be guaranteed to be accessible, someone could store the data in an unaddressable location that might violate some of the goals for that data, e.g., the creators ability to burn an NFT. This can be confusing to both the users and creators of this data.
 - Heterogeneous data cannot be stored together. In order to ensure that tokens can co-exist, the data type has been frozen, the only expressability comes via the `property_map`.
-- Data of differing types can be stored to a single data structure via `any`, but for complex data types `any` incurs additional costs within Move as each access requires deserialization. Use of `any` can also lead to run-time type mismatch as the data within an `any` can be changed throughout its lifetime.
```
davidiw, line 39: ahh, yes, I should be more explicit, that the type within an any is decided at runtime, which can impact applications which expect it to be defined. While this has no implications to safety, it can result in added logic to the indexer to make sense of types and to blockchain applications that will fail to deserialize an any if the type differs than what they expected.

I need to push out a quick PR to resolve some other differences... I'll land this and make some tweaks to better clarify.

davidiw (approved):

PR #54, opened 2023-02-01
Rename AIP files

banool:
Adding leading zeroes to the filenames so they are sorted properly. Going for 5 digits to account for up to AIP 9999.

I'm not sure if we link to these AIPs from anywhere, I figure we do, so this might break some links. But better to do it now than later.


movekevin:
> Adding leading zeroes to the filenames so they are sorted properly. Going for 5 digits to account for up to AIP 9999.
> 
> I'm not sure if we link to these AIPs from anywhere, I figure we do, so this might break some links. But better to do it now than later.

This would break a lot of links from Discord and Twitter unfortunately

banool:
We should still do it imo, because it'll be even worse if we do it later. Worth it I say to get everything nice and sorted and uniform.

Also we should avoid sharing links to GH that point at a branch instead of specific commit, to help avoid this problem.

michelle-aptos:
The links from Discord are linked to the issue so there shouldn't be a problem there

sherry-x (dismissed):
I like things properly ordered. but @michelle-aptos let's bring this up in next gatekeeper meeting to align.

banool:
I'll be out next week so if you decide to go forward with this, feel free to commandeer.

davidiw (changes requested):
we cannot simply rename, this requires a symbolic link in order to keep existing deep links.

CapCap:
For those renamed, we can keep a doc in their place that says “moved to ____”

banool:
I feel like the title is more likely to change though, and every change requires a new "redirect file". Seems a bit risky.

thepomeranian:
Hey! Thanks for your submission! I've raised this internally and, at this time, we have decided to not take action. We'll take this into consideration as we iterate and develop our AIP program!

PR #60, opened 2023-02-03
replace 'implementation' with 'deployment'

alinush:
As far as I can tell, the text in the template is asking about a **deployment** timeline on devnet, testnet and mainnet, not an _implementation_ timeline.

If it's asking about both things, let's add separate sections.

If it's asking about implementation, then I think the guidelines are confusing: i.e., what does deployment on devnet, testnet and mainnet have to do with the _implementation_ timeline?

### DIFF
--- a/TEMPLATE.md
+++ b/TEMPLATE.md
@@ -41,10 +41,14 @@ Express here the potential negative ramifications of taking on this proposal. Wh
 
 Think through the evolution of this proposal well into the future. How do you see this playing out? What would this proposal result in in one year? In five years?
 
+## Suggested implementation timeline
+
+Describe how long you expect the implementation effort to take, perhaps splitting it up into stages or milestones.
+
 ## Suggested deployment timeline
 
-When should community expect to see it on devnet
+When should community expect to see this deployed on devnet?
 
-testnet?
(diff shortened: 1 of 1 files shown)


TEMPLATE.md
```diff
@@ -41,7 +41,7 @@ Express here the potential negative ramifications of taking on this proposal. Wh
 
 Think through the evolution of this proposal well into the future. How do you see this playing out? What would this proposal result in in one year? In five years?
 
-## Suggested implementation timeline
+## Suggested deployment timeline
```
sherry-x, line 44: Let's make it "Suggested implementation and deployment timeline" instead? So far many AIPs only drafted when the code is almost complete, but for future AIPs, we would want to know when the code will be implemented and land on main branch too.

alinush: Okay. I added two separate sections: one on implementation & one on deployment, just so everything is organized neatly.

alinush: If we'd like to know when the code "lands", then I can also update the deployment timeline instructions to include an estimate of when they think the PR should land by.

sherry-x (approved):

Issue #61, opened 2023-02-04
[AIP-X][Discussion] NFT Royalty Enforcement


areshand:
# AIP Discussion

Discussion and feedback thread for AIP.

Link to AIP: https://github.com/aptos-foundation/AIPs/pull/62

## **Summary**

This document provides our recommendation and alternatives for how royalty enforcement can work on Aptos. We are collecting community feedback on the options presented here so we may decide the best path moving forward together. 

## Motivation

We observed the NFT communities on other chains are divided over the controversial topic of royalty enforcement on marketplaces. We want to ensure the creators’ royalties are respected on Aptos so creators can feel safe and incentivized to continue building on Aptos.

Currently, royalty on Aptos is dependent on our community marketplaces and dapps charging the royalties on behalf of the creators. This creates the opportunity for royalty-free marketplaces to reduce the price by skipping the royalty payments.  

Without the community forming a royalty enforcement standard together, the creators and royalty-required marketplaces are at a competitive disadvantage to royalty-free marketplaces. 

## Rationale

When comparing different options for enforcing royalties, the main considerations are:

- **The strength of the enforcement:** The enforcement should be effective enough to dissuade anyone from purchasing a token without paying royalties.
- **The impact on the future growth of the NFT projects**: The selected approach should not limit the potential of building new and exciting applications.
- **The decentralization of the approach:** The selected approach should be decentralized, agreed upon and executed by the ecosystem instead of any single entity.

### Recommended approach

**Detecting NFT Violating Royalty Treatment** 

Aptos may provide an API service that provides information to help creators find all the tokens that violate the royalty. The creator can then use the information to annotate those NFTs and flag them as royalty-violating tokens.

The API service would be built on top of an indexer that scans through the NFT transactions on-chain to look for token trades that do not pay the appropriate royalties to the creator’s royalty account. This indexer can be deployed and run by the community.

**Annotating NFT Violating the Royalty**

We can introduce a new on-chain map **RoyaltyViolation** that records all the NFTs that violate the royalty.  The creator has the authority to set the value in this on-chain map. Anyone can read from the map to check if any token has violated the royalty. 

*Note:  Only NFTs are qualified for this annotation where an NFT has a globally unique token_id.* 

```rust
Struct RoyaltyViolation has key {
	royalty_violation_token: Table<token_id, bool>
}

public bool set_royalty_violated(creator: &signer, token_id: TokenId);
public bool is_royalty_violated(token_id: TokenId);
```

[...]

PR #62, opened 2023-02-04
Add royalty enforcement AIP

areshand:


davidiw (changes requested):
While I think the overall framework makes sense, I think it is a bit ahead of itself.

* A creator can currently mutate fields if they specify so
* V2 no-code token supports this seamlessly
* it isn't clear to me that punishing the buyer for purchasing a token without a royalty is necessarily the right choice... I think there's more that needs to be studied here...

maybe it is worth converting this into an issue and leaving it there for future discussions?

areshand:
> While I think the overall framework makes sense, I think it is a bit ahead of itself.
> 
> * A creator can currently mutate fields if they specify so
> * V2 no-code token supports this seamlessly
> * it isn't clear to me that punishing the buyer for purchasing a token without a royalty is necessarily the right choice... I think there's more that needs to be studied here...
> 
> maybe it is worth converting this into an issue and leaving it there for future discussions?

There is already an issue created for future discussion if needed https://github.com/aptos-foundation/AIPs/issues/61. 

Will close this AIP for now.

PR #93, opened 2023-03-24
ed25519 validation AIP

mstraka100:
Adds an AIP "Make Ed25519 Public Key Validation Return False if Key Is the Wrong Length" in file `ed25519_pk_validation_return_none_wrong_length.md`.


aips/ed25519_pk_validation_return_none_wrong_length.md
```diff
@@ -0,0 +1,66 @@
+---
+aip: (this is determined by the AIP Manager, leave it empty when drafting)
+title: Make Ed25519 Public Key Validation Return False if Key Is the Wrong Length
+author: Michael Straka <michael@aptoslabs.com>
+Status: Draft
+type: Standard (Framework)
+created: 03/24/2023
+---
+
+# AIP-TBD - Make Ed25519 Public Key Validation Return None if Key Is the Wrong Length
```
alinush, line 10: "if key *has* wrong length"

Why The Camel Case? :)

aips/ed25519_pk_validation_return_none_wrong_length.md
```diff
@@ -0,0 +1,66 @@
...
+  
+## Summary
+
+Changes the function `native_public_key_validate` used by the native move function `native fun pubkey_validate_internal` to return False if the public key provided is the wrong length. Previously, this function would abort if the key length provided was incorrect. This change is gated by a feature flag.
```
alinush, line 14: Move, not move.

aips/ed25519_pk_validation_return_none_wrong_length.md
```diff
@@ -0,0 +1,66 @@
...
+    };
+```
+
+## Reference Implementation
+
+https://github.com/aptos-labs/aptos-core/pull/7043
+
+## Risks and Drawbacks
+
+Callers relying on the previous behavior of `native_public_key_validate` may be affected by this change.
```
alinush, line 54: Add same extra context here:

 > Currently, end-to-end tests show there are no such callers. Furthermore, it is unlikely that past callers of this function, if any, would be affected if the function returns `false` instead of aborting. This is because such callers would typically need to check the return value of the function anyway & would most likely abort anyway if the return value is `false`.

alinush (approved):

movekevin (approved):

PR #166, opened 2023-06-08
Add a Standard function to execute Move Scripts inside the Move Modules

0xAnto:
AIP to enable the execution of Move scripts from any Move modules.

### DIFF
--- a/aips/a-new-function-for-executing-move-scripts-inside-the-modules.md
+++ b/aips/a-new-function-for-executing-move-scripts-inside-the-modules.md
@@ -1,0 +1,73 @@
+---
+aip:
+title: A new Function for Executing Move Scripts inside the move modules
+author: @0xAnto
+discussions-to (*optional):
+Status: Draft
+last-call-end-date (*optional):
+type: Standard
+created: <06/08/2023>
+updated (*optional):
+requires (*optional):
+---
+
+# AIP-X - A new Function for Executing Move Scripts inside the move modules
+
+## Summary
+
+This proposal aims to introduce a new function called `execute_script` within the Move standard modules, enabling the execution of Move scripts from any move modules. By enabling this functionality, developers will have the ability to create more extensible smart contracts.
+
+## Motivation
+
+Currently, Move scripts can only be executed from the external accounts. This means that developers cannot call Move scripts from other move modules with resource accounts. This limits the extensibility of move modules. This proposal seeks to enhance the capabilities of the resource accounts for building advanced applications.
+
+## Impact
+
+This proposal impacts developers who utilize Move modules and Resource Accounts. Developers will be able to call Move scripts from Resource Accounts which can be used for the creation of smart wallets on Aptos.
+
+## Rationale
+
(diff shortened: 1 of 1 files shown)


0xAnto:
It is used for executing any functions without importing them into the module @wrwg. For example, let's consider a smart wallet module where the user should be able to execute any function. However, the current wallet module cannot achieve this because you have to import those functions into the wallet module in order to execute them. It is not practically possible to know all the functions in advance and add them to the module. Therefore, we need a way to enable users to execute any function without importing them into the module. That's why I suggest enabling script execution so that users can execute any required functions without importing them into the module.

wrwg:
I don't see that what you are asking for is related to script execution. 

Rather, it appears you like to call an arbitrary function via the web APIs, in a generic way, where you e.g. specify name of the function and list of arguments. The Aptos CLI supports this already today. We can add this to the HTTP APIs as well if it is not there yet (@banool may have an opinion/insight about this).

However, there will still be the restriction that you can only call functions which the contract allows to call from the outside. This are entry and view functions. This is needed for security of the chain.

wrwg:
Also, if you actually meant that Move code can call via some meta programming API arbitrary other Move code, than this a non-starter, sorry. This would violate a lot of the security principles build into Move, which prevent things like reentrance attacks as they are known from Eth. Move code can only call well-known other Move code, by design.

ch4r10t33r:
> Also, if you actually meant that Move code can call via some meta programming API arbitrary other Move code, than this a non-starter, sorry. This would violate a lot of the security principles build into Move, which prevent things like reentrance attacks as they are known from Eth. Move code can only call well-known other Move code, by design.

Yes, we would like for Move to support reflection. We do understand that this has some security risks. But can you pls suggest us an alternative?

wrwg:
> > Also, if you actually meant that Move code can call via some meta programming API arbitrary other Move code, than this a non-starter, sorry. This would violate a lot of the security principles build into Move, which prevent things like reentrance attacks as they are known from Eth. Move code can only call well-known other Move code, by design.
> 
> Yes, we would like to have the ability to execute any arbitrary code present in another move module. Effectively a kind of object reflection.
> 
> We do understand that this has some security risks. But can you please suggest us an alternative?

Apologies for the delay on answering, and thanks for your understanding about Move's core design principles. 

One workaround is to put your logic off-chain. Another one could be code generation: you generate Move code which dispatches to well-known call targets. I think this has been done by others before, however, this requires you to know the call targets ahead of time, and redeploy your contract the list of known target changes. Fortunately, Aptos supports code upgrade, so re-deploying is not a big deal. However, you need to keep in mind limits on code size.

Another technique I've heard mentioning is to somehow reverse the direction: instead of your generic code calling out into arbitrary other code, you let the other code call back into you. This avoids you having to hardcode the targets. However, it depends on what you plan to do, whether this works.

At some later point Move may also support a restricted, save form of dynamic dispatch. However, I still don't think the reflection like feature you are suggesting will make it into the language.

PR #175, opened 2023-06-18
Add Address Standard v1 AIP

banool:
This PR introduces an AIP for Address Standard v1. See more in the AIP.


thepomeranian (approved):

banool:
The process is even if the AIP is still a draft, we merge it in and deliberate on it elsewhere. Marking as ready for review.

PR #211, opened 2023-07-28
[AIP 47] Create aggregators_v2.md

igor-aptos:


igor-aptos:
@davidiw I can remove internal details from the AIP, but where should the "design doc" go then?

igor-aptos:
I've updated more of the text.

should we land this in the "draft" state, and then send new PRs on top of it, as needed? And potentially start external discussions soon?

thepomeranian (approved):

PR #228, opened 2023-08-11
[AIP-40] Fully specify allowed parsing behavior

banool:
Throughout the course of the AIP, those involved have discussed how parsing should work quite a lot. Generally the conversations have trended back and forth between two stances:

1. Where possible, new implementations of string -> AccountAddress functions should be as strict as possible. This tightens up how account addresses are represented in the ecosystem and ideally reduces room for error.
2. New implementations should be relaxed. If they're not, even with well thought out migration guides, we will inevitably introduce some incompatibility with a part of the ecosystem that has not / can not updated their output representation.

This PR is the synthesis of these two stances. In short, this updates the AIP to say that where possible, parsing / validation implementations should be strict, but they may also provide methods that offer more relaxed parsing. The PR updates the AIP with links to a reference implementation of this approach in the v2 TS SDK.

This allows for a clean migration. Devs who know that they will receive addresses in formats that don't comply with AIP 40 will be able to use their language's equivalent of `fromStringRelaxed`. Those who have tighter control of where input is coming from / wish to be more defensive with user input / etc. may use `fromString`. 

We have discussed this topic at length with a smaller working group so hopefully this won't be particularly contentious.


thepomeranian (approved):

0xmaayan (approved):
Looks good!

PR #230, opened 2023-08-12
[AIP-40] Update git hash used for reference implementation

banool:
I link to a specific commit rather than main so the line numbers in the URL stay correct.


banool:
Discussing offline the best way to link to code.

sherry-x (approved):

PR #233, opened 2023-08-14
add aip on safe burning of user-owned assets

davidiw:


bowenyang007 (approved):
Woohoo! No more noob noobs

aips/aip-x.md
```diff
@@ -0,0 +1,105 @@
...
+2. Deletion of user-owned objects
+  * (+) This allows a user to reclaim a refund for garbage data that likely exceeds the cost of performing the deletion, which in turn will reduce unnecessary storage load on the blockchain.
+  * (-) Some data structures cannot be deleted without potentially breaking applications on the blockchain, for example, a liquidity pool may be entirely broken if the fungible asset metadata were deleted. This could result in valuable assets being permanently locked on-chain without leveraging a governance proposal.
+
+The burn address solution was chosen as it provides a safe means to remove content associated with the users account without potentially breaking other applications.
+
+Note, the current proposal is for an `unburn` function to allow `primary_fungible_store` to reclaim accidentally transferred objects. This could instead be an alternative call into `primary_fungible_store` that explicitly has `unburn`, but that requires that wallets or other applications adopt the `unburn` functionality. The cost for checking the ownership should be negligible as the ownership must be checked prior to extracting assets.
+
+## Specification
```
movekevin, line 52: This requires thinking through other implications such as:
1. Some functions in primary_fungible_store would be in an inconsistent state - deposit would still work, withdraw/transfer would not. balance would still return the balance for an account even if that account has already "burnt" the primary store. We'd need to additionally check ownership in all of these functions and make sure they no longer work after "burn"
2. Changing balance is actually going to likely break for specific contracts such as a reward pool that has a list of owners and assumes that there are no "lost shares". With this burn functionality, shares can actually be lost now if balance suddenly returns 0 for all those owner addresses that are also separately tracked. And even more strangely, balance would still return 0 for the burn address (as the primary stores transferred through it from burn cannot be computed on chain). This effectively leads to the rewards pool suddenly losing shares. I can imagine the same problem for similar contract where the list of owners need to be updated upon transferring of any of the fungible assets representing the shares.
3. The behavior around primary stores would need to be updated as applications currently can assume if primary store exists, it's owned by the account it's created for. With "burn", this would no longer be the case.

aips/aip-x.md
```diff
@@ -0,0 +1,105 @@
...
+2. Deletion of user-owned objects
+  * (+) This allows a user to reclaim a refund for garbage data that likely exceeds the cost of performing the deletion, which in turn will reduce unnecessary storage load on the blockchain.
+  * (-) Some data structures cannot be deleted without potentially breaking applications on the blockchain, for example, a liquidity pool may be entirely broken if the fungible asset metadata were deleted. This could result in valuable assets being permanently locked on-chain without leveraging a governance proposal.
+
+The burn address solution was chosen as it provides a safe means to remove content associated with the users account without potentially breaking other applications.
+
+Note, the current proposal is for an `unburn` function to allow `primary_fungible_store` to reclaim accidentally transferred objects. This could instead be an alternative call into `primary_fungible_store` that explicitly has `unburn`, but that requires that wallets or other applications adopt the `unburn` functionality. The cost for checking the ownership should be negligible as the ownership must be checked prior to extracting assets.
```
movekevin, line 50: Are there going to be events emitted by burn/unburn?

aips/aip-x.md
```diff
@@ -0,0 +1,105 @@
...
+
+## Future Potential
+
+The future is a decentralized allow-list managed completely on-chain.
+
+## Timeline
+
+### Suggested implementation timeline
+
+This will be available before the 1.7 branch cut and available for inclusion in an earlier upgrade if the functionality is desired.
```
movekevin, line 92: Does this need to be urgently in 1.7?

davidiw: No, but I'm going to code it this afternoon, so it will be ready.

aips/aip-x.md
davidiw: Hmm... without the new event harness, I'd be concerned about adding one-off events especially for super visible action such as burn. I do reckon ... unburn is quirky without events, so let's evaluate if we really want that and if we do, I'd probably make it defined on the signer for the burn address.

aips/aip-x.md
davidiw: 1. I don't think we should make deposit fail if it is burnt. The intent of this AIP is to minimize changes to the fungible assets and digital assets. I already feel sad about the `unburn`, but I do want to ensure someone can unbreak their account if need be.
2. We could change the functions in `primary_fungible_store` but I don't think we should. Since they're primarily useful for contracts, which we intend to minimize breaking them in the first place.
3. I think we just need to update the documentation that a burned `primary_fungible_store` is still recoverable and behavior is only changed in terms of explicit ownership and those functions that depend on it (withdraw and deposit).

aips/aip-x.md
movekevin: Burn/Unburn can be introduced without events for now and we'll add events after event v2 is done

movekevin (approved):

PR #242, opened 2023-09-07
Update aip-43.md

igor-aptos:
updating AIP-43 for changes in the PRs, and addition of Fungible Asset changes

### DIFF
--- a/aips/aip-43.md
+++ b/aips/aip-43.md
@@ -1,6 +1,6 @@
 ---
 aip: 43
-title: Parallelize Digital Assets (Token V2) minting/burning
+title: Parallelize Digital Assets (Token V2) and Fungible Assets minting/burning
 author: igor-aptos (https://github.com/igor-aptos), vusirikala (https://github.com/vusirikala)
 discussions-to (*optional): https://github.com/aptos-foundation/AIPs/issues/209
 Status: In Review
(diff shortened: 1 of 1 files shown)


aips/aip-43.md
```diff
@@ -39,6 +39,20 @@ AIP 36 removed one serialization point. We can remove the rest with:
 - using Aggregators for total_supply and current_supply counters in the collection.move
 - parallelize event creation, using module events that remove sequence number altogether 
 
+In addition, we will add new API to mint tokens:
+```
+  public fun create_with_index_in_name(
```
movekevin, line 44: Is create_numbered_token a better name? cc @gregnazario the naming master

igor-aptos: open to any name :) will wait for @gregnazario 's thoughts, and update to create_numbered_token or something better

aips/aip-43.md
```diff
@@ -39,6 +39,20 @@ AIP 36 removed one serialization point. We can remove the rest with:
...
+        creator: &signer,
+        collection_name: String,
+        description: String,
+        name_with_index_preffix: String,
+        name_with_index_suffix: String,
+        royalty: Option<Royalty>,
+        uri: String,
+    ): ConstructorRef {
+```
+which allows minting of the token, where index is part of the name, while still allowing it to happen concurrently.
```
movekevin, line 54: Can we clarify whether this is a different name field from the canonical name field every token has (including those not numbered)?

movekevin (approved):

aips/aip-43.md
igor-aptos: you'll see from the other PR, but I am deprecating old name field, introducing new one to be canonical - and this method will set that one

PR #250, opened 2023-09-28
Create AIP-50: Change commission rates in delegation pools

junkil-park:


aips/aip-50.md
```diff
@@ -0,0 +1,54 @@
...
+Currently the commission rate is set by the pool owner at initiation of pool. We propose allowing the pool owner to change the commission rate after delegation pool creation.
+
+### Goals
+This feature will increase flexibility of pool owners over the network / commission rates. Pool owner can adjust operator commission rates as appropriate, based on market conditions and commercial discussions. As such, this will facilitate staking pools to participate in market dynamics. I.e. the pool can be set at a lower commission to attract more stakers or raise commission when needed. Operators can call the function themselves if they are also the pool owner, or if not, work with the pool owner they are operating for.
+
+### Out of Scope
+Operators will not be able to change commission rate. In our existing design setup, Operators are designated by pool owners.
+
+### Impact
+Delegation pool owners and corresponding operators.
```
sherry-x, line 23: we should callout the impact on delegator too, people who stake to the delegation pool previously would assume the rate is fixed and not changeable, but now they need to watch over the commission rate once for a while to make sure the operator did not overcharge them.

aips/aip-50.md
```diff
@@ -0,0 +1,54 @@
...
+
+## Alternative solutions
+There is no alternative solution because the delegation pool module currently does not allow chainging the operator commission rate.
+
+Moreover, we decided against allowing operators to change commission rate, as it would not make sense in the context of our existing design setup. Right now, Operators are designated by pool owners so eventually pool owners control everything.
+
+## Reference Implementation
+Please refer to https://github.com/aptos-labs/aptos-core/pull/10226.
```
sherry-x, line 31: Can we please also list out the high level implementation idea here for easy review? e.g. you mentioned that we will be enforcing a max commission cap, what's the value of that?

aips/aip-50.md
junkil-park: That's a good point! I added that.

aips/aip-50.md
junkil-park: Yes, I added the "specification" section where I explained what needs be implemented.

aips/aip-50.md
```diff
@@ -0,0 +1,63 @@
...
+* *Operators* are impacted by commission rate changes, which will impact their rewards.
+* *Delegators* need to watch over commission rate changes so that they are not overcharged. They may assume that the commission rate is fixed.
+
+## Alternative solutions
+There is no alternative solution because the delegation pool module currently does not allow changing the operator commission rate.
+
+Moreover, we decided against allowing operators to change the commission rate, as it would not make sense in the context of our existing design setup. Right now, Operators are designated by pool owners, so eventually, pool owners control everything.
+
+## Specification
+* The maximum commission rate is uncapped (i.e., 100%). The minimum commission rate is 0%.
```
sherry-x, line 33: when we say uncapped, it does have a max which is 100%, let's not say uncap here so people might think it can be 120% :D

aips/aip-50.md
```diff
@@ -0,0 +1,65 @@
...
+* The maximum increase rate per lockup cycle is 10%. This is to safeguard delegators from sudden changes in commission rate. For example, if the current commission rate is 10%, the new commission rate cannot be higher than 20% in the next cycle.
```
movekevin, line 34: Clarify if this is 10% absolute or 10% of current commission rate

aips/aip-50.md
```diff
@@ -0,0 +1,65 @@
...
+Please refer to https://github.com/aptos-labs/aptos-core/pull/10226.
+
+## Testing
+- What is the testing plan?
+Automated unit tests and testing in devnet/testnet before releasing to mainnet.
+
+## Risks and Drawbacks
+
+As a delegation pool accepts multiple delegators and delegators are provided the same stake-management API as the stake pool owner where `operator_commission_percentage` is taken from rewards generated, any changes to the operator commission rate will impact delegator staking rewards.
+1. We can mitigate risk to delegators by enforcing the maximum commission increase limit.
```
movekevin, line 50: In addition to this, this AIP also proposes a min remaining time in the lockup cycle for the commission rate to be updateable

aips/aip-50.md
```diff
@@ -0,0 +1,65 @@
...
+2. Emit an event when the function to update the commission rate is called. This can be tied into front-end alerts
+
+Moreover, there can be a risk that the owner changes the commission rate closer to the end of the current lockup cycle and the delegators have little time to react to the change. To mitigate this, we will allow the commission rate change request only when at least 7 days are left in the current lockup period.
+
+## Timeline
+
+### Suggested deployment timeline
+
+This change will be part of the v1.8 release.
+Testnet: December 2023.
```
movekevin, line 60: Remove this part. I think 1.8 is already sufficient

aips/aip-50.md
junkil-park: 10% absolute, according to the latest discussion with @michelle-aptos . @michelle-aptos , could you confirm this?

aips/aip-50.md
junkil-park: Fixed. Good point!

aips/aip-50.md
junkil-park: Done.

aips/aip-50.md
junkil-park: Added this. Thanks!

aips/aip-50.md
```diff
@@ -0,0 +1,72 @@
...
+    public entry fun update_commission_percentage(
+        owner: &signer,
+        new_commission_percentage: u64
+    )
+    ```
+* Only the pool owner can change the operator commission rate.
+* The maximum commission rate is 100%. The minimum commission rate is 0%.
+* The maximum absolute increase rate per lockup cycle is 10%, which is relative to the current commission rate. For example, if the current commission rate is 9%, the new commission rate cannot be higher than 19% in the next lockup cycle. This is to safeguard delegators from sudden changes in commission rate.
+* There is no maximum decrease rate.
+* The minimum remaining lockup time is 7 days. In other words, the commission rate change can be requested only when at least 7 days are left in the current lockup period.
```
alinush, line 46: This is saying that "the commission rate change can be requested only if 7 or more days are left in the current lockup period."

junkil-park: Yes, exactly. I just revised it.

aips/aip-50.md
michelle-aptos: yes absolute

michelle-aptos (approved):

aips/aip-50.md
```diff
@@ -0,0 +1,72 @@
...
+
+### Impact
+* *Delegation pool owners* have permission to change the operator commission rate.
+* *Operators* are impacted by commission rate changes, which will impact their rewards.
+* *Delegators* need to watch over commission rate changes so that they are not overcharged. They may assume that the commission rate is fixed.
+
+## Alternative solutions
+There is no alternative solution because the delegation pool module currently does not allow changing the operator commission rate.
+
+Moreover, we decided against allowing operators to change the commission rate, as it would not make sense in the context of our existing design setup. Right now, Operators are designated by pool owners, so eventually, pool owners control everything.
```
movekevin, line 32: nit: Rephrase "we decided against" to "allowing operators to change the commission rate would not make sense ..."

movekevin (approved):

aips/aip-50.md
junkil-park: Done fixing it!

PR #252, opened 2023-10-04
Create AIP-51: Changing beneficiaries for operators

junkil-park:


aips/aip-51.md
```diff
@@ -0,0 +1,65 @@
...
+
+
+### Out of Scope
+
+Changing the beneficiaries for token owners is out of the scope of this AIP.
+
+
+### Impact
+
+This feature impacts operators and their beneficiaries.
```
sherry-x, line 41: Could you please callout that the default value for beneficiary will be set to operator address? so they should expect no behavior change if they did not take any action when we release this new feature

aips/aip-51.md
```diff
@@ -0,0 +1,65 @@
...
+
+## Reference Implementation
+TBD
+
+## Testing
+- What is the testing plan?
+Automated unit tests and testing in devnet/testnet before releasing to mainnet.
+
+## Risks and Drawbacks
+Although the operator address is set to a new address, the beneficiary address could remain in the old one. The commission rewards can go to the wrong beneficiary address in this case. To prevent this, when the operator address is reset, the beneficiary should be set to the operator address.
```
sherry-x, line 52: how fast do we expect this to take effect? Is it going to be updated immediately? or it would take an epoch?

aips/aip-51.md
junkil-park: Yes, I updated accordingly. The beneficiary of any operator would be the operator itself if the operator didn't set the beneficiary explicitly.

aips/aip-51.md
junkil-park: The beneficiary change takes effect immediately. The rewards will go to the new beneficiary from the subsequent reward distribution.

Only the operator can set its beneficiary. We assume that the operator account and the beneficiary account are under the control of the operator. So, it should be fine. Please let me know if you have any concerns.

alinush (approved):

PR #254, opened 2023-10-05
Update aip-43.md - clarify breaking changes and rollout plan

igor-aptos:


aips/aip-43.md
```diff
@@ -33,6 +33,15 @@ Goal is to remove/optimize operations that make Token V2 and Fungible Asset oper
 
 This will enable higher throughput for Token V2 NFT minting/burning of a single collection, or minting/burning of a single Fungible Asset, providing better experience when there is high demand for a single collection/fungible asset.
 
+There is a breaking change for anyone accessing raw resources - like indexer or directly through the RestAPI.
```
davidiw, line 36:
```suggestion
There is a **breaking change** for anyone accessing raw resources - like indexers or directly through the RestAPI.
```

davidiw (approved):
do we wanna try to give some examples?

igor-aptos:
I kinda have to have some objects in a new way to give example links. Maybe I just manually write what it would look like. will do that separately.

PR #260, opened 2023-10-09
Create aip-54.md

movekevin:


aips/aip-54.md
```diff
@@ -0,0 +1,107 @@
...
+
+### New object_code module
+```rust
+module aptos_framework::object_code {
+  #[resource_group_member(group = aptos_framework::object::ObjectGroup)]
+  struct PublishRef has key {
+    inner: Object<PackageRegistryObject>,
+    // Extend ref of the code object.
+    extend_ref: ExtendRef,
+  }
```
lightmark, line 60: This is an object itself besides `PackageRegistryObject`?
My understanding is:
depending on whether the contract is upgradable, `extend_ref` should be an optional resource in `PackageRegistryObject`. while `PublisherRef` is unnecessary since object ownership can be used to do authorization for getting the `extend_ref`.

movekevin: Object ownership could work too. We still need to keep the ExtendRef for that case though

thepomeranian (approved):

PR #261, opened 2023-10-12
[AIP Resource Access Control] Initial document

wrwg:
This contains a draft AIP for extending the Move language by resource access control.

Related PR: https://github.com/aptos-labs/aptos-core/pull/10480

Markdown view: https://github.com/aptos-foundation/AIPs/blob/wg-rw/aips/aip-56.md

### DIFF
--- a/aips/aip-56.md
+++ b/aips/aip-56.md
@@ -1,7 +1,7 @@
 ---
-aip: 
+aip: 56
 title: Resource Access Control
-author: wg@aptoslabs.com, brian.murphy@aptoslabs.com, runtian@aptoslabs.com
+author: wg@aptoslabs.com
 discussions-to (*optional): TODO
 Status: Draft 
 last-call-end-date (*optional): <mm/dd/yyyy TBD>
@@ -9,7 +9,7 @@ type: Standard
 created: <10/11/2023>
 ---
 
-# AIP-X - Resource Access Control
+# AIP-56 - Resource Access Control
   
 ## Summary
 
--- a/aips/aip-resource-access.md
+++ b/aips/aip-resource-access.md
@@ -1,0 +1,208 @@
+---
+aip: 
+title: Resource Access Control
+author: wg@aptoslabs.com, brian.murphy@aptoslabs.com, runtian@aptoslabs.com
+discussions-to (*optional): TODO
+Status: Draft 
+last-call-end-date (*optional): <mm/dd/yyyy TBD>
+type: Standard
+created: <10/11/2023>
+---
+
+# AIP-X - Resource Access Control
+  
+## Summary
+
+In Move, functions can access arbitrary resources as long as they have access to public APIs allowing to manipulate the resource. There are multiple problems with this, as discussed in the [Motivation](#Motivation) section. This AIP proposes an extension to the Move language which allows for fine-grained access control of resources. This is achieved by generalizing the familiar `acquires T` declaration in Move in a downwards compatible way. The evaluation of the resulting access control discipline is primarily _dynamic_, but intended to become static in the future. The dynamic-first approach is natural because the redundancy principle of "paranoid" VM mode makes a dynamic check necessary anyway.
+
+### Goals
+
+This AIP intends to achieve the following:
+
+- Have an organic extension of the Move language to support access control, which is fully downwards compatible
+- Enable new strategies for parallelization and sharding
+- Increase confidence in the effects of transactions and functions for users and auditors by access control declarations
+- Enable safe dynamic dispatch by making access control part of a type.
+
+### Out of Scope
+
+This feature will only be supported in the upcoming Aptos Move compiler ("compiler v2").
+
+Static analysis of access control will initially not be implemented, but is expected to be ready when AMC gets out of beta.
+
+## Motivation
+
+A Move function can read, write, and create arbitrary resources as long as it has access to the module APIs which allow this, and/or is in possession of a signer. There are multiple problems with this:
+
+- It has been identified in the past as a security risk. For example, a transaction that obtains a
+  signer and then delegates to some other well-known contract -- a 'helper' wrapper offered to some user -- can create arbitrary resources, including deploying code, without the user's knowledge, under the users account.
+- Unspecified read/write behavior of transactions makes it harder to parallelize for block execution and to implement sharding strategies.
+- It is difficult to reason about side effects of functions and transactions. This is both the case for auditors as for formal analysis like the Move prover. The only way to reason is to look at the code.
+- Since looking at code is the only way to reason about effects, _calling unknown_ code (dynamic dispatch across trust boundaries) is unsafe, and therefore currently forbidden in Move.
+
+## Impact
+
+Significant impact is expected in the medium term on two frontiers: on the one hand sharding and parallelization, on the other enabling other new language features like dynamic dispatch. 
+
+
+## Alternative solutions
+
+In Solana, a transaction must declare upfront the accounts it reads or writes from. Similarily, in Sui Move a transaction declares what objects it accesses. Using Aptos Objects as basis for what is proposed here has been discussed. Arguably, the approach described here is closer to the core Move language, as it does not depend on library extensions like objects. The approach is also more general, as it allows a variety of different access patterns, including negation.
+
+
+## Specification
+
+### Source Language
+
+The existing `acquires T` notation in Move is replaced by an extended syntax:
+
+```move
+fun f() acquires R { .. }                      // today: reads or write at any address; only local to module
+fun f() reads M1::R writes M2::T { .. }        // new access kinds -- note reference to types from other modules
+fun f<T> acquires R<T> { .. }                  // type instantiations supported
+fun f() acquires 0x42::* { .. }                // wildcards
+fun f(x: T) acquires *(address_of(x)) { .. }   // data dependency
+fun f() acquires !0x1::* { .. }                // negation
+fun f() pure { .. }                            // pure function, no accesses
+```
+
+If a function has multiple access clauses the interpretation is as follows: positive (non-negated) clauses build a union, and negative clauses an intersection. This in `reads A !writes B reads C !writes D` access is allowed if it is either a read of `A` or of `B`, _and_ if it is not a write of C and D.
+
+### File Format
+
+The existing `FunctionHandle` is extended by a field which contains an optional list of access specifiers. Each access specifier consists of a kind (read/write/..), a resource specifier, and an address specifier. Both those data types are chosen to represent the various forms of patterns as seen in the source language. For details, see the implementation.
+
+### Semantics
+
+#### Access Specifier Subsumption
+
+The conceptual syntax of access specifiers is as follows:
+
+```
+AccessSpecifiers := { AccessSpecifier } 
+AccessSpecifier := [ ! ] Kind ResourceSpecifier AddressSpecifier
+ResourceSpecifier := * | Address::* | Address::Module::* | Address::Module::Resource [ TypeArgs ]
+AddressSpecifier := * | Address | Parameter | Function Parameter
+Kind := acquires | reads | writes
+```
+
+The basic function specifying the semantics of access specifiers is _subsumption_ (denoted as `>=` here). One list of access specifier subsumes another one if all access allowed in the other are also allowed in the subsuming one. For `ResoureSpecifier` this looks as below. Notice that transitivity of subsumption is assumed:
+```
+*         >=     _
+A::*      >=     A::M::*
+A::M::*   >=     A::M::R
+A::M::R   >=     A::M::R<TS>
+```
+
+Subsumption for `AddressSpecifier` is similar but depends on an environment to obtain values for parameters and specific functions for them. Assume this environment is called `E`:
+
+```
+*         >=     _
+x         >=     y       iff E(x) == E(y)
+f(x)      >=     g(y)    iff E(f(x)) == E(g(y))
+```
+
+Notice that the functions which are allowed in address specifiers are selected from a small set of well-known builtins, like e.g. obtaining the address of a signer or of an Aptos object. No arbitrary Move code is executed.
+
+Given the subsumption, one can now define a function to test whether given pair of resource and address is covered by an access specifier:
+
+```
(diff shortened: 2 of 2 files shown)


aips/aip-resource-access.md
```diff
@@ -0,0 +1,196 @@
...
+last-call-end-date (*optional): <mm/dd/yyyy TBD>
+type: Standard
+created: <10/11/2023>
+---
+
+# AIP-X - Resource Access Control
+  
+## Summary
+
+In Move, functions can access arbitrary resources as long as they have access to public APIs allowing to manipulate the resource. There are multiple problems with this, as discussed in the [Motivation](#Motivation) section. This AIP proposes an extension to the Move language which allows for fine-grained access control of resources. This is achieved by generalizing the familiar `acquires T` declaration in Move in a downwards compatible way. The evaluation of the resulting access control discipline is primarily _dynamic_, but intended to become static in the future. The dynamic-first approach is natural because the redundancy principle of "paranoid" VM mode makes a dynamic check necessary anyway.
```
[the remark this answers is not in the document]
brmataptos: Also, is "paranoid" VM mode documented somewhere?  If not, it could perhaps use more explanation here.

aips/aip-resource-access.md
```diff
@@ -0,0 +1,196 @@
...
+
+### File Format
+
+The existing `FunctionHandle` is extended by a field which contains an optional list of access specifiers. Each access specifier consists of a kind (read/write/..), a resource specifier, and an address specifier. Both those data types are chosen to represent the various forms of patterns as seen in the source language. For details, see the implementation.
+
+### Semantics
+
+#### Access Specifier Subsumption
+
+The conceptual syntax of access specifiers is as follows:
```
brmataptos, line 76: Your syntax is lacking "!" (negation) and you don't have any way to express union (multiple paths).   Perhaps there is a higher-level comma-separated AccessSpecifierList which conveys union?

aips/aip-resource-access.md
```diff
@@ -0,0 +1,196 @@
...
+
+### Source Language
+
+The existing `acquires T` notation in Move is replaced by an extended syntax:
+
+```move
+fun f() acquires R {}                      // today: reads or write at any address; only local to module
+fun f() reads M1::R writes M2::T {}        // new access kinds -- note reference to types from other modules
+fun f<T> acquires R<T> {}                  // type instantiations supported
+fun f() acquires 0x42::*                   // wildcards
```
brmataptos, line 63: You should include a body here and in the next line, or this is confusing.  Might be useful to bold-face the access specifier in the example to distinguish it from the other function spec syntax ("fun f() ... {}").

aips/aip-resource-access.md
```diff
@@ -0,0 +1,196 @@
...
+*         >=     _
+A::*      >=     A::M::*
+A::M::*   >=     A::M::R
+A::M::R   >=     A::M::R<TS>
+```
+
+Subsumption for `AddressSpecifier` is similar but depends on an environment to obtain values for parameters and specific functions for them. Assume this environment is called `E`:
+
+```
+*         >=     _
```
brmataptos, line 97: I guess (from knowledge of Rust) that you mean * >>= anything, but I think you should be more explicit here.

aips/aip-resource-access.md
```diff
@@ -0,0 +1,196 @@
...
+
+The existing `acquires T` notation in Move is replaced by an extended syntax:
+
+```move
+fun f() acquires R {}                      // today: reads or write at any address; only local to module
+fun f() reads M1::R writes M2::T {}        // new access kinds -- note reference to types from other modules
+fun f<T> acquires R<T> {}                  // type instantiations supported
+fun f() acquires 0x42::*                   // wildcards
+fun f(x: T) acquires *(address_of(x))      // data dependency
+fun f() acquires !0x1::* {}                // negation
```
brmataptos, line 65: To declare a pure function, do I need to say:
```
fun f() acquires !* {}
```
That seems a bit awkward.  Perhaps some clearer syntax meaning "nothing" would be useful.

aips/aip-resource-access.md
```diff
@@ -0,0 +1,196 @@
...
+
+Subsumption for `AddressSpecifier` is similar but depends on an environment to obtain values for parameters and specific functions for them. Assume this environment is called `E`:
+
+```
+*         >=     _
+x         >=     y       iff E(x) == E(y)
+f(x)      >=     g(y)    iff E(f(x)) == E(g(y))
+```
+
+Notice that the functions which are allowed in address specifiers are selected from a small set of well-known builtins, like e.g. obtaining the address of a signer or of an Aptos object. No arbitrary Move code is executed.
```
brmataptos, line 102: Can we explicitly relate this to some of the Aptos Object operators here?

Actually, I see a weakness of this formulation w.r.t. Aptos Objects.  The whole object model presumes that addresses created by well-known builtins like create_object_from_account() do not alias anything else.  Given that, we should be able to assume that addresses either provided explicitly or created by other means will not alias addresses from create_object_from_account().  Can we represent this usefully?  I have a function that can won't be anything generated by create_object_from_account(*) nor will it be certain other functions.  Perhaps something like `literal(*)` == addresses disjoint from the range of any builtin function.

aips/aip-resource-access.md
```diff
@@ -0,0 +1,196 @@
...
+fun f(x: T) acquires *(address_of(x))      // data dependency
+fun f() acquires !0x1::* {}                // negation
+```
+
+### File Format
+
+The existing `FunctionHandle` is extended by a field which contains an optional list of access specifiers. Each access specifier consists of a kind (read/write/..), a resource specifier, and an address specifier. Both those data types are chosen to represent the various forms of patterns as seen in the source language. For details, see the implementation.
+
+### Semantics
+
```
brmataptos, line 73: I suggest an overview of semantics, perhaps:

The `reads`, `writes`, and `acquires` access specifiers for a function respectively represent a *superset* of the global resources which may be read, written, or read and/or written by that function.  The current (Aptos) Move Book states that 

> A Move function m::f must be annotated with acquires T if and only if:
> - The body of m::f contains a move_from<T>, borrow_global_mut<T>, or borrow_global<T> instruction, or
> - The body of m::f invokes a function m::g declared in the same module that is annotated with acquires

but this proposal loosens that requirement.  To allow for future changes, a function may declare `acquires`, `reads`, or `writes` to resources that are not currently accessed by the function.  But if a function accesses a resource, then it must have been declared.

(Not sure how to make that backwards compatible, though.)

aips/aip-resource-access.md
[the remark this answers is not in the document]
wrwg: > "downwards"? Perhaps you mean "backwards".

They are synonyms. You can Google it.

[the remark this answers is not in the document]
wrwg: > Also, is "paranoid" VM mode documented somewhere? If not, it could perhaps use more explanation here.

We have blogs about paranoid mode ("Securing Move"), so yes I think this is sufficiently well known.

aips/aip-resource-access.md
wrwg: Done.

aips/aip-resource-access.md
wrwg: Done

aips/aip-resource-access.md
wrwg: There is now a `pure` specifier and I added it as an example.

aips/aip-resource-access.md
wrwg: I'm not sure whether I get the point. Object address can be aliased. In `fun f(o1: Object<T>, o2: Object<T>)` the both parameters can be the same or different objects. The creation is a different story, but this model is not for creation, but for read/write.

aips/aip-resource-access.md
wrwg: I added something more to the semantics section, but this is intentional vague here, its still under development. We update it once things finalize.

aips/aip-resource-access.md
brmataptos: The case that is interesting is this: can we determine that a function which reads/writes an "object" definitely does not conflict with a function that reads/writes a non-object resource?  Or do we assume that the Resource type will be different for those two cases so that can be used to disambiguate?

For example, here is some  [example code I found](https://aptos.dev/standards/aptos-object/) referring to objects:
```
public entry fun modify_reserves(liquidity_pool: Object<LiquidityPool>) {
    let liquidity_pool = &mut borrow_global_mut<LiquidityPool>(liquidity_pool);
    liquidity_pool.reserves_a = liquidity_pool.reserves_a + 1000;
}
```
What are the appropriate access specs for this function?

brmataptos (commented):
More comments after reading the code PR.  I now realize the code is partially documented here, but that's not very findable for users of that code.

We should pull all past AIP docs into the Move Book, and comment our code appropriately.

aips/aip-resource-access.md
wrwg: Access spec for this function would be `writes LiquidityPool(object::owner(liquidity_pool))`. Note that this example has a bug. You can only call borrow on addresses. It should have been

```
    let liquidity_pool = &mut borrow_global_mut<LiquidityPool>(object::owner(liquidity_pool));
``` 

In general, I don't think you have a way to determine non-overlapping accesses, unless with a complex reasoning machine. Specifically at runtime, this is not possible. In fact when this function is called, `writes LiquidityPool(object::owner(liquidity_pool))` will be simplified based based on the concrete parameter to `writes LiquidityPool(0x42)`.

wrwg:
> More comments after reading the code PR. I now realize the code is partially documented here, but that's not very findable for users of that code.
> 
> We should pull all past AIP docs into the Move Book, and comment our code appropriately.

There is an extended section about the semantics in `runtime_access_specifier.rs`. That is the best place for it since the evaluation logic is also in this file. Commenting all the various ASTs seems to be less useful and duplication, as the same concept has multiple instances through the phases of the compiler.

Indeed the Move book will have added a complete new section about access specifiers which explains semantics, gives examples, etc., in a style understandable for users. Its different than this AIP, but maybe some stuff can be pulled from here. However, we should add it to the Move book not before we are closer to get this into production.

aips/aip-resource-access.md
```diff
@@ -0,0 +1,208 @@
...
+fun f() reads M1::R writes M2::T { .. }        // new access kinds -- note reference to types from other modules
+fun f<T> acquires R<T> { .. }                  // type instantiations supported
+fun f() acquires 0x42::* { .. }                // wildcards
+fun f(x: T) acquires *(address_of(x)) { .. }   // data dependency
+fun f() acquires !0x1::* { .. }                // negation
+fun f() pure { .. }                            // pure function, no accesses
+```
+
+If a function has multiple access clauses the interpretation is as follows: positive (non-negated) clauses build a union, and negative clauses an intersection. This in `reads A !writes B reads C !writes D` access is allowed if it is either a read of `A` or of `B`, _and_ if it is not a write of C and D.
+
```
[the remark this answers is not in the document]
wrwg: Done

aips/aip-resource-access.md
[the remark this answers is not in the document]
wrwg: I think it comes from my 10 years at Microsoft, they often use different notions than the mainstream. BTW I can happily change this if it is better English (which I'm certainly no expert in), I was just defending that the notion in fact exists as a synonym.

aips/aip-56.md
```diff
@@ -0,0 +1,213 @@
...
+
+Note that unless trust boundaries are concerned (transactions and public functions), we rarely expect users to write access specifiers, so the stack is not expected to become very deep in average. 
+
+#### Gas Cost
+
+Access specifiers need to be metered, also to prevent their misuse. This is different from how Aptos currently treats paranoid mode. Because this feature adds additional utility to users, metering is justified.
+
+#### Compatibility
+
+Since access specifiers are encoded in the `FunctionHandle` instances which are copied from imported modules, they get outdated when the dependency is upgraded. The compatibility rule for upgrade is that access specifiers can only be constrained, that is `S_v1 >= S_v2 >= ...`
```
vineethk, line 141: Not sure if this is something to bring up in the AIP discussion necessarily, but it might be worth writing some guidance and considerations about compatibility (things to be careful about, maybe suggested compatibility patterns) when it comes to access control specifiers. In particular, negation along with wildcards disallows forever accessing structs that don't even exist right now (which is okay, and perhaps even desired in many cases, but we should say more about compatibility rules when we roll this feature out).

If I originally wrote some code like so:

```
module a {
  fun foo() !reads a::* {}
}
```

And later wanted to upgrade the code like so:
```
module a {
  struct A {}
  struct B {}
  fun foo() reads A, !reads B {}
}
```

It would be incompatible, because `A` is in the updated specifier, but not in the original specifier (even when `A` is a non existent entity in the original specifier).

I realized one could use "wrappers" to split up types into allow lists or ban lists in perpetuity, as generic types give a new "namespace" in the resource access specifier language.

```
module a {
  struct DisAllow<T> { wraps: T }
  fun foo() !reads DisAllow {}
}
```

In future versions, one can add new structs to the module, and as long as they are not wrapped with `DisAllow`, they can be read by `foo()` and its callees.

aips/aip-56.md
```diff
@@ -0,0 +1,213 @@
...
+See [PR #10544](https://github.com/aptos-labs/aptos-core/pull/10544) for the VM implementation.
+
+## Testing
+
+The behavior of access specifiers can be well tested with unit tests. Significant coverage is expected. 
+
+
+## Risks and Drawbacks
+
+Risks specifically include developer adoption and engineering complexity. This feature need to be well documented with intuitive examples. We may also build tools which allow to derive access specifiers for public functions automatically, easing the effort.
```
vineethk, line 157: If it turns out that multiple functions are being annotated with the same (long list of) access specifiers, we could consider adding an "access specifier alias" (i.e., an explicitly declared name for a list of access specifiers, that could be used in lieu of the full list).

igor-aptos (approved):

aips/aip-56.md
wrwg: This is interesting, perhaps pull this out into an issue?

aips/aip-56.md
wrwg: Yes. Though I would expect only public functions are annotated, the rest inferred.

PR #273, opened 2023-11-09
Update aip-53.md

davidiw:


aips/aip-53.md
```diff
@@ -49,7 +49,7 @@ In the current model, everyone knows who the fee payer is. This gives a false se
...
+It is possible that two fee payers could compete for paying for a transaction. The current mempool design will only support a single variant at a time. Therefore a malicious entity could intercept the transaction and replace it with a faulty fee payer. As a result, the transaction could be discarded in consensus resulting in a form of censorship. While the likelihood is small due to the inherent trust model and network dynamics, this still remains a risk. As a result, application developers are recommended to specify the fee payer when possible rather than leveraging the optional fee payer model.
```
alinush, line 52: "inherent trust model and network dynamics" is rather hand-wavy.

If it's not easy to state what is sufficient/necessary, than maybe give some examples:. e.g., sending your TXN to malicious PFN is sufficient for your TXN to be DoS'd.

aips/aip-53.md
```diff
@@ -49,6 +49,8 @@ In the current model, everyone knows who the fee payer is. This gives a false se
 
 This does, however, empower the user to generate a single transaction that can then be wrapped by distinct fee payers. From a visible perspective, this is not entirely different than what exists today. In either case, the Aptos Mempool design prevents two transactions with the same sequence number from being registered or executed, hence there is no consequence that comes from this change as extra transactions will be discarded prior to charging the fee payer.
 
+It is possible that two fee payers could compete for paying for a transaction. The current mempool design will only support a single variant at a time. Therefore a malicious entity could intercept the transaction and replace it with a faulty fee payer. As a result, the transaction could be discarded in consensus resulting in a form of censorship. While the likelihood is small due to the inherent trust model and network dynamics, this still remains a risk. As a result, application developers are recommended to specify the fee payer when possible rather than leveraging the optional fee payer model.
```
davidiw, line 52:
```suggestion
It is possible that two fee payers could compete for paying for a transaction. The current mempool design will only support a single variant at a time. Therefore a malicious entity could intercept the transaction and replace it with a faulty fee payer. As a result, the transaction could be discarded in consensus resulting in a form of censorship. While the likelihood is small due to the inherent trust model and network dynamics, this still remains a risk. Specifically, transactions are typically submitted to fullnodes trusted to forward the transaction to validator fullnodes, who in turn have the same expectation -- setting an invalid fee payer and forwarding is more work than discarding the transaction. As a result, application developers are recommended to specify the fee payer when possible rather than leveraging the optional fee payer model.
```

alinush (approved):

PR #278, opened 2023-12-05
Digital Assets Composability

Aladeenb:
A no-code solution to achieve composability securely using digital and fungible assets.


davidiw:
Thanks for your contribution! The team has been wrapping up the release 1.9 over the past week and hasn't had time yet to review this. I, personally, should have some free time next week.

Aladeenb:
Thanks for the reply David! Looking forward to that :)

Aladeenb:
Link to the source code: 

https://github.com/TowneSquare/TowneSpace-contract/blob/dev/sources/core.move

If there is any better way to include this in the AIP PR, feel free to let me know.

davidiw (commented):
Overall I like the sentiment that we can benefit from more no-code solutions.

I think I'd have to review the code to really understand what is proposed here, so this AIP feels a bit incomplete.

The other thing I'd like to think about is that there are a few issues with aptos-token, in particular, it wasn't built to be extended, which means that any other no-code solutions have to be rebuilt with potential copying of code, maybe that's not such a big issue.

I'm not sure I agree on the extra cost of keeping around "unnecessary" refs or why you feel those refs are unnecessary.

Also the security implications aren't very clear, how and why this is superior, it would be good to lay out the claims with maybe an example.

Finally, I think the right way for us to start thinking about standards in this space isn't to publish more under governance addresses, but to figure out how to take the core work in governance addresses and extend it outside. Then the AIPs are more about saying here's the canonical implementation and it can be found at address X on network Y (0xabcd on mainnet). That way you get more time to evolve this with ecosystem involvement, it doesn't have to ever be published into a less flexible domain, such as governance addresses. Governance addresses are a pain to publish to and upgrade due to the very nature that they require broad agreement.

Aladeenb:
> The other thing I'd like to think about is that there are a few issues with aptos-token, in particular, it wasn't built to be extended, which means that any other no-code solutions have to be rebuilt with potential copying of code, maybe that's not such a big issue.

We have plans on copying some of the features and functions from aptos-token, this will be worked on along the way as the main focus now is  the composability feature.

>  I'm not sure I agree on the extra cost of keeping around "unnecessary" refs or why you feel those refs are unnecessary.

These are initially deemed unnecessary based on our needs. I have revised my opinion and considering building building a no-code solution that has aptos-token's features, I have realised the inverse. 

>  Also the security implications aren't very clear, how and why this is superior, it would be good to lay out the claims with maybe an example.

Updating the URI Should Be a Scarce Operation

Consider a scenario where you want to compose or decompose two digital assets, A and B.

1. Using Framework Modules:

- Compose/Decompose digital assets using the transfer function from the object module: For composing, transfer B to A; for decomposing, send B back to the owner's wallet.

- The native transfer functions do not automatically update the URI associated with token A. This is a separate, mandatory step required to reflect the updated composition.

- Composed digital assets should not be transferred separately without updating the URI. Failure to update the URI will result in a misleading representation of the asset, making it difficult for users to visualise the actual composition. This opens up opportunities for malicious actors to deceive users into believing a digital asset possesses more assets than it actually does.

- If asset B possesses refs, they must also be transferred alongside it so that the new owner of B is the only one that can use those refs.

2. Using Our Composability Module:

- Compose/Decompose digital assets using the transfer function from `core.module`: For composing, transfer B to A; for decomposing, send B back to the owner's wallet.

- During both operations, an argument for the updated uri have to be inputted. This process is now manual but we're not intended to keep it this way.

- B refs are stored within the same object as the B, meaning they're sticked to B.

I will add an example later to illustrate better the differences mentioned.

> Finally, I think the right way for us to start thinking about standards in this space isn't to publish more under governance addresses, but to figure out how to take the core work in governance addresses and extend it outside. Then the AIPs are more about saying here's the canonical implementation and it can be found at address X on network Y (0xabcd on mainnet). That way you get more time to evolve this with ecosystem involvement, it doesn't have to ever be published into a less flexible domain, such as governance addresses. Governance addresses are a pain to publish to and upgrade due to the very nature that they require broad agreement.

I agree with this.

Aladeenb:
Hi @davidiw,

I'd like to update you on the recent developments regarding the proposal, taking into account the suggestion you provided.

Here are the key highlights:

- The refined solution now incorporates a more expansive architecture, offering not only a no-code solution but also a customisable option. This is achieved by returning the `constructor_ref` from the internal creation functions instead of `object`. Creators interested in a simplified approach to composability can directly engage with the no-code solution, specifically `studio.move`, where functions output and require objects rather than constructor references.
- Leveraging `Aptos-token.move`: In response to your suggestion, we've incorporated the code from `Aptos-token.move`. This enhancement provides creators with broader options when creating and updating collections and tokens.

You can find the updated proposal [on v0.2.0-alpha branch](https://github.com/TowneSquare/TowneSpace-contract/tree/v0.2.0-alpha) and simple example implementation [here](https://github.com/TowneSquare/TowneSpace-contract/blob/v0.2.0-alpha/examples/sources/hero.move). Your additional insights on these improvements would be greatly appreciated.

Thank you!

Aladeenb:
Hi @davidiw,

I've incorporated the changes we discussed for the proposal in the last pull requests. The goals are now explicitly defined, and I've made the necessary adjustments. 

Feel free to take a quick look at the latest version. Thank you!

aips/aip-x.md
```diff
@@ -0,0 +1,625 @@
+---
+aip:
```
davidiw, line 2: 76 is free

davidiw (commented):
I think it is worthwhile to land this after the few recommended tweaks. This doesn't signify acceptance or approval, but it is a well written proposal that deserves a more permanent place. Once you wrap this up, please feel free to respond to the PR or message me on telegram.

Aladeenb:
Hey @davidiw , the PR should be ready now.

davidiw (approved):

PR #279, opened 2023-12-05
[AIP-X] create APT fungible asset at 0x1

lightmark:
Create APT FA at 0x1.


lightmark:
https://github.com/aptos-foundation/AIPs/pull/280
covered this

PR #296, opened 2024-01-13
OpenID-based blockchain (OIDB) accounts

alinush:


magnum6actual:
This is absolutely huge. We've been toying around with ways of trying to do this.  100% support this AIP concept.

alinush:
> This is absolutely huge. We've been toying around with ways of trying to do this. 100% support this AIP concept.

What a coincidence! We've been doing some toying around too! 😁

sherry-x (approved):

PR #305, opened 2024-01-26
create wallet standard aip

0xmaayan:


aips/wallet-standard.md
```diff
@@ -0,0 +1,325 @@
...
+last-call-end-date (*optional): 09/02/2024
+type: Standard
+created: 29/01/2024
+---
+
+# AIP-X - Wallet Standard
+
+## Summary
+
+A Wallet Standard is an API interface a wallet implements and defines how applications interact with a wallet. This AIP introduces a new wallet standard for the Aptos ecosystem.
```
davidiw, line 16:
```suggestion
The Wallet standard defines a universal API for wallet and application interactions. This AIP introduces a new wallet standard for the Aptos ecosystem.
```

davidiw (commented):
as far as I can tell, this doesn't conflict with supporting the original (an non-standardized) flow?

0xmaayan:
> as far as I can tell, this doesn't conflict with supporting the original (an non-standardized) flow?

correct. If a wallet still supports the original standardized flow it should work as long as the dapp that interacts with it supports it also. If a wallet supports the new standardized flow then the dapp should support the new one also.

At the end, this is just a set of APIs and interfaces that follow a new (and better) standard.

aips/wallet-standard.md
```diff
@@ -0,0 +1,325 @@
...
+getAccount():Promise<UserResponse<AccountInfo>>
+```
+
+`aptos:getNetwork` to get the current network in the wallet
+
+```ts
+getNetwork(): Promise<UserResponse<NetworkInfo>>;
+```
+
+`aptos:signAndSubmitTransaction` method to sign and submit a transaction using the current connected account in the wallet.
```
hardsetting, line 93: I think this should be optional.
The same functionality can be achieved by an `aptos:signTransaction` + submitting the transaction using the AptosClient.

The new Aptos Wallet Adapter can easily provide a fallback implementation.

aips/wallet-standard.md
```diff
@@ -0,0 +1,325 @@
...
+
+```ts
+const onConnect = () => {
+  this.wallet.features["aptos:connect"].connect();
+};
+```
+
+## Risks and Drawbacks
+
+The new standard uses the [new TypeScript SDK](https://github.com/aptos-labs/aptos-ts-sdk) types and therefore requires dapps and wallets to use/migrate to the new TypeScript SDK or hold a conversion layer from the new TypeScript SDK types to the legacy TypeScript SDK types.
```
hardsetting, line 304: Just throwing this out there, but it'd be really cool to have a migration tool, that automatically converts your code to use the new SDK.

I know it's not trivial though.

aips/wallet-standard.md
```diff
@@ -0,0 +1,325 @@
...
+export type AptosSignMessageOutput = {
+  address?: string
+  application?: string
+  chainId?: number
+  fullMessage: string
+  message: string
+  nonce: string
+  prefix: 'APTOS'
+  signature: Signature | Signature[]
+  bitmap?: Uint8Array
```
hardsetting, line 188: I think this should be just `Signature`.
Both `MultiEd25519Signature` and `MultiKeySignature` already contain a bitmap field.

aips/wallet-standard.md
```diff
@@ -0,0 +1,325 @@
...
+}
```
hardsetting, line 189: I think one of the first priorities after getting this standard out would be to work on a message signature standard, that includes domain separation and allow structured messages (we've been getting lots of requests about this).
Alin should probably be pulled for this

hardsetting (approved):

aips/wallet-standard.md
0xmaayan: agreed. the adapter (and I believe also Petra plugin) already uses this kind of a conversion layer, we can get it into its own package to be used by wallets

aips/wallet-standard.md
0xmaayan: I also think a wallet should only supports `sign` functionality. no `signAndSubmitTransaction` or `submitTransaction`. will make this optional for a wallet to implement

aips/wallet-standard.md
```diff
@@ -0,0 +1,325 @@
...
+
+A standard feature is a method that must or should be supported and implemented by a wallet.
+Here is a list of the suggested [Aptos features](https://github.com/aptos-labs/wallet-standard/tree/main/src/features)
+
+> a feature marked with `*` is an optional feature
+
+`aptos:connect` method to establish a connection between a dapp and a wallet.
+
+```ts
+// `silent?: boolean` - gives ability to trigger connection without user prompt (for example, for auto-connect)
```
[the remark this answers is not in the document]
0xmaayan: I believe this is for autoConnect functionality. maybe the user has given the wallet permission to auto connect and doesnt want to aprrove the popup everytime. Same feature as one of the wallets have - i think it is Fewcha?

davidiw (approved):
address @hardsetting 's feedback :pray:

aips/wallet-standard.md
0xmaayan: @hardsetting do you mean we should remove the `bitmap` prop or change its type to `Signature`?

hardsetting: I'm guessing `Signature` here is the abstract class / union type for _any_ signature
i.e. `Ed25519Signature | MultiEd25519Signature | AnySignature | MultiKeySignature`

since `MultiEd25519Signature` and `MultiKeySignature` already contain a field for `bitmap` and `signatures`, I would change this like such:

```suggestion
  signature: Signature
```

aips/wallet-standard.md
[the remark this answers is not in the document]
hardsetting: oh I see, wasn't aware we wanted to allow that

thepomeranian (approved):

PR #317, opened 2024-02-09
[AIP-66] Passkey Accounts

hariria:


aips/passkeys.md
```diff
@@ -0,0 +1,504 @@
...
+- [**WebAuthn Authenticator**](https://www.w3.org/TR/webauthn-3/#authenticator): A cryptographic entity, existing in hardware or software, that can register a user with a given relying party and later assert possession of the registered public key credential, and optionally verify the user to the relying party.
+
+
+### Aptos Account Creation with Passkeys
+
+During the registration of a new WebAuthn credential, an asymmetric key (i.e., a **private key** and its associateed **public key**) is generated securely on that device. The public key associated with the WebAuthn credential can be used to derive an associated account on chain. The private key, which controls the account, will be secured by the device.
+
+The registration process is particularly important for several reasons:
+
+1. The public key associated with the credential is only revealed during registration. Assertion responses (signing with the passkey) do NOT return the public key. Thus, the registration response must be parsed and the relevant information about the credential should be stored properly and persisted in storage by the relying party.
```
alinush, line 80: This will need to be addressed. Otherwise, it sounds like if you lose your **public** key, you lose your account, which is no bueno.

alinush: (Good news is ECDSA supports [pubkey recovery](https://cryptobook.nakov.com/digital-signatures/ecdsa-sign-verify-examples): given a message and a signature, you can extract a set of up to 2 candidate pubkeys.)

alinush: But this does mean that adopting passkeys for other signature schemes will **NOT** be a good idea unless this is addressed.

aips/passkeys.md
```diff
@@ -0,0 +1,504 @@
...
+
+To make a WebAuthn credential compatible with Aptos, the `pubKeyCredParams` array should only contain supported signature schemes. At the present moment, only `secp256r1` is supported, which requires that the array have one element with `alg: -7` to denote `secp256r1` as the choice of credential type.
+
+**[`PublicKeyCredentialUserEntity`](https://www.w3.org/TR/webauthn-3/#dictdef-publickeycredentialuserentity)**
+
+Each authenticator stores a [credentials map](https://www.w3.org/TR/webauthn-3/#authenticator-credentials-map), a map from ([`rpId`](https://www.w3.org/TR/webauthn-3/#public-key-credential-source-rpid), [[`userHandle`](https://www.w3.org/TR/webauthn-3/#public-key-credential-source-userhandle)]) to public key credential source. If the user creates another credential with the same user handle as an existing credential on that authenticator, it will **OVERWRITE** the existing credential, thus overwriting the private key associated with a passkey account. To avoid this, it is **IMPERATIVE** for the relying party to maintain a list of previously registered credentials and corresponding user handles.
```
alinush, line 117: It is not helpful to say relying party here, since it ignores the context in which this is deployed. Instead, clarify who this is: the dapp, the wallet, and the implication on users. 

e.g., Could a user be screwed over by a buggy wallet or dapp?

alinush: I think it would help to clarify that "creates another credential [...] on that authenticator" means "on that device/platform". Maybe give an example in parantheses.

aips/passkeys.md
```diff
@@ -0,0 +1,504 @@
...
+#### Operating Systems
+
+See the [operating system reference](https://passkeys.dev/docs/reference/) to better understand what operating systems support passkeys and their limitations. Most notably, passkeys registered on Windows devices will NOT be backup-able as of February 2023. If the user would like the passkey to be backed up, it is recommended that they use a device that allows for multi-device credentials and supports passkey backup.
+
+#### Browsers
+
+See this [device support matrix](https://passkeys.dev/device-support/) for more info on which browsers and operating systems enable backup-able passkeys.
+
+### Wallet liveness
+
```
alinush, line 474: The discussion here is once again not very helpful, you need to talk about impact on dapps, wallets, and not relying parties.

aips/passkeys.md
```diff
@@ -0,0 +1,504 @@
...
+In other words, if the relying party is unavailable or returns an error like a 404, the user will not be able to use the passkey associated with that relying party until the relying party comes back online.
+
+To learn more about how chromium handles assertion and registration requests, see [`webauthn_handler.h`](https://github.com/chromium/chromium/blob/95bb60bf7fd3d18f469f050b60663b3dbdfa0402/content/browser/devtools/protocol/webauthn_handler.h#L19)
+
+### Compromised cloud provider
+
+If your passkey is a multi-device credential, backed up to iCloud or Google Password Manager, and the associated cloud account is compromised, so is your passkey account.
+
+Note this only affects multi-device credentials, not hardware-bound credentials like Yubikeys as those are not backed up.
+
```
alinush, line 488: You did not talk about two important risks you mentioned before:
 - Losing the PK means losing the account (and mitigations via ECDSA pubkey recovery)
 - "If the user creates another credential with the same user handle as an existing credential on that authenticator, it will **OVERWRITE** the existing credential"
 
 Both of these need to be addressed.

aips/passkeys.md
```diff
@@ -0,0 +1,504 @@
...
+**[`PublicKeyCredentialUserEntity`](https://www.w3.org/TR/webauthn-3/#dictdef-publickeycredentialuserentity)**
+
+Each authenticator stores a [credentials map](https://www.w3.org/TR/webauthn-3/#authenticator-credentials-map), a map from ([`rpId`](https://www.w3.org/TR/webauthn-3/#public-key-credential-source-rpid), [[`userHandle`](https://www.w3.org/TR/webauthn-3/#public-key-credential-source-userhandle)]) to public key credential source. If the user creates another credential with the same user handle as an existing credential on that authenticator, it will **OVERWRITE** the existing credential, thus overwriting the private key associated with a passkey account. To avoid this, it is **IMPERATIVE** for the relying party to maintain a list of previously registered credentials and corresponding user handles.
+ 
+**[`AuthenticatorSelectionCriteria`](https://www.w3.org/TR/webauthn-3/#dictdef-authenticatorselectioncriteria)**
+
+This field defines additional information about the credential and provides constraints that affect the UX of registering / asserting with the credential. See comments above the fields below for more info:
+
+```js
```
alinush, line 123: This pseudocode and comments are not helpful. The AIP should translate this W3C standard gobbledygook for the reader & describe implications for the actual use case (Aptos/blockchains).

aips/passkeys.md
```diff
@@ -0,0 +1,504 @@
...
+Every WebAuthn credential is created by the relying party (RP) with a set of options, called `PublicKeyCredentialCreationOptions`. These options help configure many aspects of the WebAuthn credential, including, but not limited to:
+
+- The asymmetric key type for the credential (`Ed25519`, `ECDSA`, `RSA`, etc.)
+- User verification requirements during assertions
+- Whether the WebAuthn credential should be bound to one device or available cross-platform. 
+
+It's important to carefully consider which options to select as they have significant impacts on the passkey user experience and **CANNOT** be reconfigured after credential creation.
+
+```
+dictionary PublicKeyCredentialCreationOptions {
```
alinush, line 94: Not sure if this level of detail is necessary.

aips/passkeys.md
```diff
@@ -0,0 +1,504 @@
...
+| extensions             | variable (if present)     | Extension-defined authenticator data. This is a CBOR [RFC8949] map with extension identifiers as keys, and authenticator extension outputs as values. See § 9 WebAuthn Extensions for details.                                                                                                                                                                               |
+
+
+The `authenticatorData` in an attestation response (`AuthenticatorAttestationResponse`) contains `flags` that provide information on the **backup eligibility** and **backup state** of the WebAuthn credential. If the relying party (RP) deems that backup should be required for a WebAuthn credential, the RP should check the Backup Eligibility (`BE`)and Backup State (`BS`) flags in the `AuthenticatorAttestationResponse` to ensure that the credential is backed up. Having both `BE` and `BS` set to true implies that the [credential is a multi-device credential and is currently backed up](https://www.w3.org/TR/webauthn-3/#sctn-credential-backup). See the rust representation of the flags below:
```
alinush, line 239: Again, translate relying party in term of cryptocurrency usecases...

aips/passkeys.md
```diff
@@ -0,0 +1,504 @@
...
+
+1. Let `options` be an instance of [`PublicKeyCredentialCreationOptions`](https://www.w3.org/TR/webauthn-3/#dictdef-publickeycredentialcreationoptions), configured to the [Relying Party](https://www.w3.org/TR/webauthn-3/#relying-party)'s needs for the ceremony.
+2. Call [`navigator.credentials.create()`](https://w3c.github.io/webappsec-credential-management/#dom-credentialscontainer-create) and pass options as a parameter to the function.
+
+#### Registration Response
+
+After successfully calling [`navigator.credentials.create()`](https://w3c.github.io/webappsec-credential-management/#dom-credentialscontainer-create), the authenticator will return a [`PublicKeyCredential`](https://www.w3.org/TR/webauthn-3/#publickeycredential). In the `PublicKeyCredential` will be a `response` field of type [`AuthenticatorAttestationResponse`](https://www.w3.org/TR/webauthn-3/#authenticatorattestationresponse) which will contain most of the information needed for verification of the registration response.
```
alinush, line 153: Again, not sure that this level of detail is at all necessary.

aips/passkeys.md
```diff
@@ -0,0 +1,504 @@
...
+
+With the raw public key, an Aptos account address can be derived. Note that the default authentication key derivation will vary depending on the account type created (`SingleKey` or `MultiKey`).
+
+### Signing a transaction
+
+In a traditional client-server implementation of the WebAuthn specification, both registering a new credential and authentication assertion use [challenge-response authentication](https://en.wikipedia.org/wiki/Challenge%E2%80%93response_authentication) to avoid several types of attacks. Randomized challenges are particularly important for protecting against [replay attacks](https://en.wikipedia.org/wiki/Replay_attack) [(§13.4.3)](https://www.w3.org/TR/webauthn-3/#sctn-cryptographic-challenges).
+
+Aptos is adapting the WebAuthn specification to be used for on-chain accounts and transactions. As many already know, on-chain transactions include a sequence number to avoid replay attacks on Aptos. Thus, if transactions are used as the challenge in a WebAuthn assertion, the replay attack risk is mitigated even if the challenge is not randomly generated by the relying party.
```
alinush, line 299: This kind of explanation is very nice.

aips/passkeys.md
```diff
@@ -0,0 +1,504 @@
...
+
+#### [`challenge`](https://www.w3.org/TR/webauthn-3/#authenticatorassertionresponse)
+
```
alinush, line 302: Again, not sure if this level of detail is necessary.

aips/passkeys.md
```diff
@@ -0,0 +1,504 @@
...
+    sequence<PublicKeyCredentialDescriptor> allowCredentials = [];
+    DOMString                               userVerification = "preferred";
+    sequence<DOMString>                     hints = [];
+    DOMString                               attestation = "none";
+    sequence<DOMString>                     attestationFormats = [];
+    AuthenticationExtensionsClientInputs    extensions;
+};
+```
+
+During an assertion, the `SHA3-256` digest of `signing_message(raw_transaction)` is provided as the `challenge` in [`PublicKeyCredentialRequestOptions`](https://www.w3.org/TR/webauthn-3/#dictdef-publickeycredentialrequestoptions). The [`signing_message`](https://github.com/aptos-labs/aptos-core/blob/main/crates/aptos-crypto/src/traits.rs#L151) function is applied before computing the `SHA3-256` of the `raw_transaction` to acheive domain separation without further increasing the size of the challenge.
```
alinush, line 317: It's good that you explain that passkeys are used to sign over a challenge and we simply replace that challenge with a non-replayable TXNs.

Again, not sure that you need to explain in low level all of the details.

alinush (commented):
I think the implementation details are too low-level for an AIP. I do like that you have them, since they serve as good reference for us & others later on. I just proposed they be moved to an appendix. This way the AIP can focus on:
1. How we use passkeys **as a black box** to bootstrap passkey-based blockchain accounts
2. The **implications** of doing this, given how this black box is actually implemented (e.g., might lack back up, PK might go missing, passkey might be overwritten, etc). Note that you do not need to list W3C structs or our Rust structs to drive these points across.

aips/passkeys.md
hariria: Thanks for the suggestion. For simplicity, I will clarify in `terminology` that "relying party" is synonymous with a wallet when discussing passkey accounts.

aips/passkeys.md
hariria: Removed

aips/passkeys.md
hariria: Removed

aips/passkeys.md
hariria: Removed the `Registering WebAuthn Credential` section

aips/passkeys.md
```diff
@@ -0,0 +1,412 @@
...
+created: <12/14/2023>
+updated (*optional): <mm/dd/yyyy>
+requires (*optional): <AIP number(s)>
+---
+
+# AIP-X - Passkey accounts
+
+## Summary
+
+This AIP proposes the first WebAuthn Authenticator for Aptos, enabling users to utilize passkeys and other WebAuthn credentials for on-chain accounts and transaction signing.
```
davidiw, line 18:
```suggestion
This AIP proposes the first WebAuthn Authenticator for Aptos, enabling users to utilize passkeys and other WebAuthn credentials for transaction authentication.
```

aips/passkeys.md
```diff
@@ -0,0 +1,412 @@
...
+
+[Passkeys](https://fidoalliance.org/passkeys/) are designed to replace passwords as a phishing resistant, faster, and more secure form of user authentication. When a user registers a passkey, a new [public key credential](https://www.w3.org/TR/webauthn-3/#public-key-credential) is created on their device's [authenticator](https://www.w3.org/TR/webauthn-3/#authenticator) - which holds and manages the credentials for that user. As of February 2023, on iOS 16+, Android 9+, and MacOS 13+, the passkey can then securely sync with their device's cloud provider (iCloud or Google Password Manager, for example) using end-to-end encryption to ensure that the passkey is recoverable in the event that the device is lost. When signing in with the passkey, the user is prompted to authenticate - often with a biometric based [authorization gesture](https://www.w3.org/TR/webauthn-3/#authorization-gesture) like Face ID or Touch ID - and provides a response that proves ownership of the passkey via a digital signature.
```
davidiw, line 20:
```suggestion
[Passkeys](https://fidoalliance.org/passkeys/) are designed to replace passwords as a phishing resistant, faster, and more secure form of user authentication. When a user accesses a pass-key enabled site, it asks the user to register a passkey specific for that site. When a user registers a passkey, a new [public key credential](https://www.w3.org/TR/webauthn-3/#public-key-credential) is created on their device's [authenticator](https://www.w3.org/TR/webauthn-3/#authenticator). In future sessions with that website, the passkey can be used in lieu of a password to produce a digital signature that validates the identity of the user. Authenticators store passkeys behind the users device credentials, which can be accessed via biometric-based [authorization gesture](https://www.w3.org/TR/webauthn-3/#authorization-gesture) like Face ID or Touch ID.
```

This paragraph is saying too much.

Maybe talk about recovery in the open problems space rather than here:

" As of February 2023, on iOS 16+, Android 9+, and MacOS 13+, the passkey can then securely sync with their device's cloud provider (iCloud or Google Password Manager, for example) using end-to-end encryption to ensure that the passkey is recoverable in the event that the device is lost."

aips/passkeys.md
```diff
@@ -0,0 +1,412 @@
...
+
+Passkeys use the [WebAuthn specification](https://www.w3.org/TR/webauthn-3/), a standard co-created by the [FIDO alliance](https://fidoalliance.org/) and the [World Wide Web Consortium (W3C)](https://www.w3.org/), for passkey registration and authentication. We are choosing to use NIST P256 (`secp256r1`) as the first enabled WebAuthn signature scheme because of its broad support across most modern operating systems. Once complete, this will enable users to sign and submit transactions with any compatible WebAuthn credential on Aptos, including multi-device credentials registered on iOS, MacOS, and Android devices, as well as single-device, hardware-bound credentials on devices like Yubikeys.
```
davidiw, line 22: Probably too much content, let's try to rewrite this purely as a summary for application to Aptos: 

```suggestion
On Aptos, passkeys authenticate transactions by using a [WebAuthn](https://www.w3.org/TR/webauthn-3/)-specific `AccountAuthenticator`. `AccountAuthenticators` are mentioned in [AIP-55](https://github.com/aptos-foundation/AIPs/blob/main/aips/aip-55.md). Aptos currently supports NIST P256 (`secp256r1`) as the only WebAuthn signature scheme because of its broad support across most modern operating systems. This functionality allows a users to sign and submit transactions with any compatible WebAuthn credential on Aptos, including multi-device credentials registered on iOS, MacOS, and Android devices, as well as single-device, hardware-bound credentials on devices like Yubikeys.
```

aips/passkeys.md
```diff
@@ -0,0 +1,412 @@
...
+
+## Motivation
+
+Blockchains have revolutionized digital asset ownership by providing users with the ability to fully control their account without a centralized custodian. This decentralized model, however, has drawbacks. If the user self-custodies their keys, they are fully responsible for the management of their account and there is no recovery path if users lose their private key.
+
+Passkeys provide a great choice for users to create secure private keys that are recoverable in the event of device loss. Furthermore passkeys make onboarding more seamless by enabling a user to create an account without having to write down a plaintext mnemonic or private key.
```
davidiw, line 28:
```suggestion
Passkeys offer a seamless onboarding flow into Web3 as users can create and recover an account without relying own storing plaintext mnemonics or private keys. Specifically, passkeys within Apple or Android ecosystems can easily be recovered or migrated to new devices. Further opportunities exist within Aptos with its native key rotation and support for K-of-N multikey accounts, as discussed in AIP-55.
```

aips/passkeys.md
```diff
@@ -0,0 +1,412 @@
...
+
+- To enable users on Aptos to create an on-chain account associated with their WebAuthn credential
+- To enable users on Aptos to sign transactions with a WebAuthn credential
+
+## Impact
+
+This AIP will benefit developers and users by providing them with an easy way to create non-phishable, recoverable private keys.
+
+1. **User-friendliness:**
+   1. WebAuthn credential registration (private key creation) and transaction signing can be done simply via device biometrics
```
davidiw, line 42:
```suggestion
   1. WebAuthn credential registration (private key creation) and transaction signing can be done simply via user authentication such as device biometrics
```

aips/passkeys.md
davidiw: This depends on the signature though, doesn't it?

davidiw: Feels like we really ought to have account creation and account rotation function that take in a vector of public keys so that the authentication key can be verified and each public key stored as an event...

aips/passkeys.md
```diff
@@ -0,0 +1,412 @@
...
+
+On the other hand, there are tradeoffs to consider as well:
+
+1. There are centralization and liveness concerns both on the OIDC provider and the services that OIDB accounts require to function in a privacy preserving manner (e.g., pepper service, proving service) respectively.
+2. Unless the ephemeral secret key ($\mathsf{esk}$) associated with the OIDB account of the user is encrypted (i.e., with a password) or is a passkey itself, the plaintext ephemeral secret key is still available on the browser, leaving it potentially vulnerable to a malicious actor (but only during its short expiration window). Passkey private keys, on the other hand, are stored on the device when the passkey is backed up, often within a secure element on the device and not on the browser.
+
+## Specification
```
davidiw, line 65: Could we have a very brief overview of the end to end flow?

* User visits site and is asked to register a passkey for the site
* User generates an aptos address from the passkey using ...
* the user transfers funds to that account or the dapp offers sponsored transactions (AIP-52) to create the account
* the user initiates an action within a dapp
* the dapp asks for permissions to submit a transaction
* the system prompt asks for permissions to use the passkey
* the passkey signs the user transaction using webauthn

aips/passkeys.md
hariria: There's a bit of nuance here since the authenticator is not always the device itself, it could be a password manager for example, so I will clarify this while removing the unnecessary content

aips/passkeys.md
hariria: Will word-smith and add

aips/passkeys.md
hariria: Added

davidiw (approved):

aips/aip-66.md
```diff
@@ -0,0 +1,442 @@
+---
+aip: 66
+title: Passkey Accounts
+author: hariria
+discussions-to (*optional): <a url pointing to the official discussion thread>
+Status: <Draft | Last Call | Accepted | Final | Rejected>
```
sherry-x, line 6: status is draft

aips/aip-66.md
```diff
@@ -0,0 +1,442 @@
+---
+aip: 66
+title: Passkey Accounts
+author: hariria
+discussions-to (*optional): <a url pointing to the official discussion thread>
```
sherry-x, line 5: link to the discussion issue

PR #342, opened 2024-02-20
amend aip-64

zjma:


thepomeranian (approved):

davidiw (commented):
I think one thing that seems relevant here is that a ValidatorTransaction implies that the validators may have different states and one validator is sharing their perspective... otherwise perhaps there would be a different framework, where there's just an extension to the existing block metadata?

PR #343, opened 2024-02-20
aip 67 update

zjma:
Calling out the assumptions of when JWK consensus will work and examples of when it will not.


thepomeranian (approved):

zjma:
> thanks for giving a consideration on how it might get stuck, could you specify what the ramifications of getting stuck... which is eventually the process will timeout (next epoch?) and should get to eventual consistency... we believe this to be unlikely enough that we will have yet to implement more sophisticated error handling at this point in time... or something like that?

@alinush correct.me if I'm wrong, but
jwk consensus is a best-effort approach and there's no real ramification by design. And I believe all oidc-based applications won't work if such inconsistency happens...

PR #390, opened 2024-03-23
Aptos Token Objects Framework Update AIP

JohnChangUK:
This AIP aims to extend the `collection`, `token` and `property_map` modules from the `aptos-token-objects` framework. Reasons for this are to introduce extensibility and customization to collection and token objects.

Details in AIP


aips/aip-x.md
```diff
@@ -0,0 +1,158 @@
...
+
+This AIP aims to extend the `collection`, `token` and `property_map` modules from the `aptos-token-objects` framework. Reasons for this are to introduce extensibility and customization to collection and token objects.
+
+## Motivation
+
+Currently there is no way of changing the name or max supply of a collection, nor create a token with the token and custom seed provided. Reasons for wanting this is for example:
+
+1. Some users may not want to have on chain data publicly visible, such as the collection name during collection creation. Being able to set this at a later date allows users to confidently deploy collections without leaking any sensitive information.
+2. Users may want to adjust the max supply for a collection. This can occur if there needs to be adjustments in total supply, or simply rectifying mistakes.
+3. If a user wants to predict the address of a token before creating this on chain, this AIP would allow this by taking in the token and custom seed. From this, the address of the token can be known beforehand. An example of why this is needed is preparing metadata for tokens that are yet to be minted.
```
davidiw, line 26: `create_named_token` already provides this, what are we gaining by adding the additional here?

the motivating example is already solved with the current standard.

JohnChangUK: What if I want the same token name for all tokens in a collection, as well as predictable addresses? The current implementation would fail as it takes the `name` as the seed. Users may want this option.

davidiw: add it as a motivating example :P

aips/aip-x.md
```diff
@@ -0,0 +1,158 @@
...
+
+## Summary
+
+This AIP aims to extend the `collection`, `token` and `property_map` modules from the `aptos-token-objects` framework. Reasons for this are to introduce extensibility and customization to collection and token objects.
+
+## Motivation
+
+Currently there is no way of changing the name or max supply of a collection, nor create a token with the token and custom seed provided. Reasons for wanting this is for example:
+
+1. Some users may not want to have on chain data publicly visible, such as the collection name during collection creation. Being able to set this at a later date allows users to confidently deploy collections without leaking any sensitive information.
```
davidiw, line 24: I was reminded why this isn't so great and needs more thought:
* https://github.com/aptos-labs/aptos-core/blob/5a82310b900d65257edb7969a358580bc2af453b/aptos-move/framework/aptos-token-objects/sources/token.move#L247
* https://github.com/aptos-labs/aptos-core/blob/5a82310b900d65257edb7969a358580bc2af453b/aptos-move/framework/aptos-token-objects/sources/token.move#L268

by enabling the users to move the name, we kinda break collection accesses and have to remember a legacy name. It is a bit weird... thoughts on correctness and usability if we do that? do we want to add a few more functions that allow the devs to be explicit on collection address?

bowenyang007: That's the assumption to begin with as when we first created you told us that the name could be changed. We built indexing with this in mind and I'm pretty sure that sdk reads from indexer.

Also 99.9% collections won't do this. The worst case is that the collections that do do this will have to figure out how to manage their own collection id.

davidiw: This is a risk that needs to be explicitly called out in this doc, in the code, and our documentation. Also it is great if the indexer tracks the original collection name and we add APIs that enable us to create new tokens given a collection id.

davidiw: If we give something really risky, we should have clarity on how users can manage this in a reasonable fashion.

JohnChangUK: Token creation functions which take in `Object<Collection>` will be created, as this solves this issue as discussed!

aips/aip-x.md
JohnChangUK: Added

aips/aip-x.md
```diff
@@ -0,0 +1,217 @@
...
+2. Users may want to adjust the max supply for a collection. This can occur if there needs to be adjustments in total supply, or simply rectifying mistakes.
+3. If a user wants to predict the address of a token before creating this on chain, this AIP would allow this by taking in the token and custom seed. From this, the address of the token can be known beforehand. An example of why this is needed is preparing metadata for tokens that are yet to be minted.
+    1. Solidity provide this mechanism with contract deployment, via both **[CREATE](https://docs.openzeppelin.com/cli/2.8/deploying-with-create2#create)** and [**CREATE2**](https://docs.openzeppelin.com/cli/2.8/deploying-with-create2) opcodes. This enables users to send funds to an address before the smart contract has been deployed to the address.
+4. Users may want the same token name for all tokens in a collection, as well as predictable addresses. The current implementation would fail as it takes the name as the seed.
```
davidiw, line 28:
```suggestion
```

davidiw (approved):
minor cleanups to merge the same idea

feel free to send another PR if you want to make tweaks

PR #392, opened 2024-03-26
[aip63] revise case study

lightmark:


davidiw (commented):
what are we storing on chain to ensure case 5? we need some additional resource on the PFA that specifies that this was a result of a migration, right?

lightmark:
Yes, so I add the `MigrationFlag` with a simple diagram.

davidiw (approved):

PR #399, opened 2024-04-09
randomness aip sepcification update

zjma:


aips/on-chain-randomness-impl.md
```diff
@@ -148,11 +127,211 @@ In each epoch, the validators will collaborate to produce randomness for *every*
...
+The wVRF evaluation of needs to be put on chain as the block randomness seed in the `BlockMetadata` transaction (always the 1st transction in a block).
+To be backward-compatible, a new transaction type `BlockMetadataExt` is needed (and updating `BlockMetadata` transaction is impossible).
+```rust
+pub enum Transaction { // the internal transaction type
+    UserTransaction(SignedTransaction),
+    GenesisTransaction(WriteSetPayload),
+    BlockMetadata(BlockMetadata),
+    StateCheckpoint(HashValue),
+    ValidatorTransaction(ValidatorTransaction),
+    BlockMetadataExt(BlockMetadataExt), // new variant
```
davidiw, line 145: how will the chain know to use this new variant and then to switch to v1?

zjma: it's feature gated [here](https://github.com/aptos-labs/aptos-core/blob/1de391c3589cf2d07cb423a73c4a2a6caa299ebf/consensus/src/state_computer.rs#L203), also called out in a later section

PR #410, opened 2024-04-23
AIP-82: Transaction Context Extension

junkil-park:


e99243506bigplay:
Could we change gas payer to specific signer in smart contract? so sender no gas but still able to execute this function, paid by smart contract or someone else.

davidiw (approved):

davidiw:
> Could we change gas payer to specific signer in smart contract? so sender no gas but still able to execute this function, paid by smart contract or someone else.

I think this belongs more as a feature request on aptos-core but also belongs to a more generic account abstraction model that's in the works.... good request.

e99243506bigplay:
> > Could we change gas payer to specific signer in smart contract? so sender no gas but still able to execute this function, paid by smart contract or someone else.
> 
> I think this belongs more as a feature request on aptos-core but also belongs to a more generic account abstraction model that's in the works.... good request.

ah sorry, I mean, include conditional changing gas payer that if smart contract is successful to execute then wont cost gas otherwise transaction failed will sender cost that gas amount.

davidiw:
> > > Could we change gas payer to specific signer in smart contract? so sender no gas but still able to execute this function, paid by smart contract or someone else.
> > 
> > 
> > I think this belongs more as a feature request on aptos-core but also belongs to a more generic account abstraction model that's in the works.... good request.
> 
> ah sorry, I mean, include conditional changing gas payer that if smart contract is successful to execute then smart contract will cost gas otherwise transaction failed will sender cost that gas amount.

It might be good to give an example, but again this is a behavioral change, this AIP is purely about exposing internal information about the transaction to the VM.

thepomeranian (approved):

PR #414, opened 2024-04-27
framework-level untransferable fungible asset stores

davidiw:
@lightmark

### DIFF
--- a/aips/aip-82.md
+++ b/aips/aip-82.md
@@ -1,0 +1,149 @@
+---
+aip: 82
+title: Framework-level Untransferable Fungible Asset Stores
+author: davidiw, 
+discussions-to (*optional): <a url pointing to the official discussion thread>
+Status: Draft
+type: Framework
+created: 04/27/2024
+updated (*optional): <mm/dd/yyyy>
+requires: 21
+---
+
+# AIP-82 - Framework-level Untransferable Fungible Asset Stores
+  
+## Summary
+
+For better control over assets on-chain providers must be able to completely freeze an address from accessing any of the provided asset and not just the existing fungible stores owned by the address.
+
+While the assets within a fungible asset stores can be frozen via `fungible_asset::set_frozen_flag`, that does not prevent the owner of the asset from creating new fungible stores and continue to access assets. This AIP introduces a new `fungible_asset::create_untransferable_store` API that ensures that asset providers can restrict access and offers users of the Aptos framework a unified API to generate all of their asset stores.
+
+### Out of Scope
+
+This is not intended to be a completely general purpose solution as this is not extensible for applications outside of a pure fungible asset store. For example, some asset providers have used fungible asset stores, paired with digital assets to create various new concepts like asset time-locked assets and fungible digital assets. Due to the nature of the required restrictions, this targets only the specific application where a fungible asset store should never be moved.
+
+# High-level Overview
+
+During the creation of a new fungible asset via `add_fungibility`, the creator can also call `set_untransferable`, which will indicate that the underlying `create_store` function cannot be called by external applications and that the only way to create stores is via a new API: `create_untransferable_store`. The store created from `create_untransferable_store` will have the `object::transfer` method revoked to prevent transfers.
+
+If at any point in time, the creator decides to freeze an account, they need only freeze the primary fungible account, creating and freezing one, if it does not exist.
+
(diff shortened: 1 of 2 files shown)


aips/aip-82.md
```diff
@@ -0,0 +1,151 @@
...
+author: davidiw, 
+discussions-to (*optional): <a url pointing to the official discussion thread>
+Status: Draft
+type: Framework
+created: 04/27/2024
+updated (*optional): <mm/dd/yyyy>
+requires: 21
+---
+
+# AIP-82 - Framework-levle Untransferable Fungible Asset Stores
```
flex-alcove, line 13: seems a typo `level`

aips/aip-82.md
```diff
@@ -0,0 +1,149 @@
...
+/// Set that only untransferable stores can be create for this fungible asset.
+public fun set_untransferable(constructor_ref: &ConstructorRef) {
+    let metadata_addr = object::address_from_constructor_ref(constructor_ref);
+    assert!(exists<Metadata>(metadata_addr), error::not_found(EFUNGIBLE_METADATA_EXISTENCE));
+    let metadata_signer = &object::generate_signer(constructor_ref);
+    move_to(metadata_signer, Untransferable {});
+}
+
+/// Creates an untransferable store.
+public fun create_untransferable_store<T>(account: &signer, metadata: Object<T>): Object<FungibleStore> {
```
lightmark, line 80: We should allow the creator to customize this object so the signature should be similar to `create_store`. So I believe the signature should be:
`public fun create_untransferable_store<T>(constructor_ref: ConstructorRef, metadata: Object<T>): Object<FungibleStore>`
Note now we move `ConstructorRef` into this function to ensure it cannot be used later to reset any flag.

aips/aip-82.md
```diff
@@ -0,0 +1,149 @@
...
+    let constructor_ref = object::create_object_from_account(account);
+    
+    // Freeze the object
+    let obj_transfer_ref = object::generate_transfer_ref(&constructor_ref);
+    object::disable_ungated_transfer(&obj_transfer_ref);
+    
+    if (object::can_generate_delete_ref(&constructor_ref)) {
+        let delete_ref = object::generate_delete_ref(&constructor_ref);
+        let store_signer = &object::generate_signer(&constructor_ref);
+        move_to(store_signer, Deletable { delete_ref });
```
lightmark, line 90: not related: do we want to do this as well for normal fungible store?

aips/aip-82.md
```diff
@@ -0,0 +1,149 @@
...
+    };
+    create_store_internal(&constructor_ref, metadata)
+}
+
+/// Creates a store on an object during construction.
+public fun create_store<T: key>(
+    constructor_ref: &ConstructorRef,
```
lightmark, line 97: if we start with `ConstructorRef` instead of `&ConstructorRef`, this current solution would be much cleaner.

I propose to deprecate this function and use a new `create_fungible_store` function that takes `ConstructorRef` in a unified way and ban transfer based on whether `Untransferable` exists. The reason is for projects dealing with FA you don't want them to do 
```
if (it is transferrable) { create_store } else {create_untransferrable_store}... The api call should be transparent to users.

File: aips/aip-82.md [hunk 5]
```diff
@@ -0,0 +1,149 @@
...
+
+### Out of Scope
+
+This is not intended to be a completely general purpose solution as this is not extensible for applications outside of a pure fungible asset store. For example, some asset providers have used fungible asset stores, paired with digital assets to create various new concepts like asset time-locked assets and fungible digital assets. Due to the nature of the required restrictions, this targets only the specific application where a fungible asset store should never be moved.
+
+# High-level Overview
+
+During the creation of a new fungible asset via `add_fungibility`, the creator can also call `set_untransferable`, which will indicate that the underlying `create_store` function cannot be called by external applications and that the only way to create stores is via a new API: `create_untransferable_store`. The store created from `create_untransferable_store` will have the `object::transfer` method revoked to prevent transfers.
+
+If at any point in time, the creator decides to freeze an account, they need only freeze the primary fungible account, creating and freezing one, if it does not exist.
```
lightmark, line 29: I have the same idea before reading this line :)

aips/aip-82.md
davidiw: If we allow for `ConstructorRef`, we have no way of enforcing the non-transfer aspect.

aips/aip-82.md
davidiw: No, the idea for `create_store` is that the developer can decide what they want to do.... because this is self-contained, we need to be more thoughtful

aips/aip-82.md
davidiw: Same issue, I can generate a `TransferRef` then pass in the `ConstructorRef` and bypass any logic here.

movekevin (approved):

PR #415, opened 2024-04-29
api and on-chain config updates for randomness

zjma:
- API should be discussed in a separate AIP (AIP-41?)
- updated on-chain conig refactoring specifications


thepomeranian (approved):

thepomeranian (approved):

alinush:
So the undergasing attack prevention stuff was removed because... it should be in AIP-41? There is already something there, but nothing too concrete (e.g., always require 2APT).

zjma:
@alinush it was added back with more details in #417

PR #431, opened 2024-05-27
AIP: Canonical Image Templates for Token Collections

SaulDope:
Our team at mirage protocol is building a web2 server to generate dynamic images for Tokens based on their underlying data.  I started thinking on how we could do it purely on-chain and came up with this.

The idea is still new and I'm very open to feedback.  Looking forward to discussing this.


SaulDope:
Here's some examples of others who've attempted similar:

### [OpenSea](https://docs.opensea.io/docs/metadata-standards#metadata-structure)

They allow users to upload an entire SVG as `image_data` inside NFT metadata, but they say they don't recommend this.  This approach definitely works but quickly brings up storage concerns as there would be one image stored on-chain per NFT.


### [Fully On-Chain Art](https://www.fullyonchain.art/articles/fully-on-chain-svg-nfts-part-2-erc-721-nft-contract)
This organization promotes EVM based fully on-chain art.  ERC-721 has more flexibility in the on-chain metadata, allowing for inline SVGs.  They do a lot more of the SVG building on-chain than I think is needed, but one idea I really appreciated was putting a pure function to render the SVG on-chain.  This helps unify client side approaches.  Pure functions in EVM don't require knowledge of on-chain state, so they can be fully ported to the browser to avoid reads/view function calls to node operators.  

The core idea of the approach in Fully On-Chain has similarities to my proposal above.  However the implementation is a lot more work for developers, primarily because they're (to some degree) trying to force it into the standard.  Parsing SVGs raw in Move is something I am hesitant about which is why my proposal did not include it.  However it should at least be considered, so I'll continue to think on it.

PR #436, opened 2024-05-30
AIP-83 clarification

igor-aptos:


igor-aptos:
Fix is being cherry-picked, so AIP-83 will provide full guarantees from the time its released,  no need for text adjustment

PR #448, opened 2024-06-11
Create AIP for Aptos Script/Intent Builder

runtian-zhou:


ying-w:
In order to differentiate between (legacy) scripts and newly introduced batched execution, could you force these transactions to emit an event that also has information on what entry functions and arguments are being batched

aips/aip-x.md
```diff
@@ -0,0 +1,128 @@
...
+updated (*optional): <mm/dd/yyyy>
+requires (*optional): <AIP number(s)>
+---
+
+# AIP-X - Batched Execution Transaction
+  
+
+## Summary
+
+Right now the users can interact with a Move contract via entry point function or script. Script can be a nice way of chaining multiple smart contract calls yet require users to compile scripts on their own end, which would be a problem for a light weight client setup.
```
davidiw, line 19: We should build on the motivation here.

The existing paradigm for developing on Aptos requires end developers to contemplate all the ways in which a user may interact with their application. This has many implications such that an application cannot introduce support for new asset classes without being updated and that interactions between libraries must be defined completely on-chain. For example, objects allow for ungated transfers and transfers with `TransferRef`; however, a developer cannot seamlessly introduce asset specific transfers with `TransferRef` without support from applications. Applications then need to tailor their support around these assets. FungibleAsset addressed this by introducing dispatchable function, but this does not lend itself to a general solution. Secondly, as a user, an `Entry` function is opaque and there is no way to be explicit from the user about the intent of their transaction, such that, I want to extract `5 apt` from my account to acquire `Aptomingo X`. Instead a user must review simulation of the transaction and hope that the simulation outcome is reasonably deterministic.

This AIP introduces mechanisms that push interactions on the blockchain to be a transaction generation time decision to address these concerns and make user and developer lives on Aptos better by making it easier to build applications and safely determine what operations will be performed.

aips/aip-x.md
```diff
@@ -0,0 +1,151 @@
...
+        function: `0x1::coin::withdraw`,
+        functionArguments: [BatchArgument.new_signer(0), 1],
+        typeArguments: ["0x1::aptos_coin::AptosCoin"]
+    });
+
+    /* return_2 should contain the converted FungibleAsset */
+    let return_2 = await builder.add_batched_calls({
+        function: `0x1::coin::coin_to_fungible_asset`,
+        /* Pass the withdrawed Coin<AptosCoin> to the coin_to_fungible_asset */
+        functionArguments: [return_1[0]],
```
igor-aptos, line 93: it seems a bit odd to assume tuple is returned, and require [0] always.
would it be better to have return_1 by default, and if it is tuple you can specify return_1.tuple_index(0) or something?

igor-aptos (approved):
can you add a bit more about the motivation with respect to onchain vs dynamic composability, to address the David's comment

After that this looks good to land, and iterate.

PR #450, opened 2024-06-13
Create AIP for order votes

vusirikala:
Adding a new AIP for order votes


aips/aip-order-votes.md
```diff
@@ -0,0 +1,226 @@
...
+	if |pending_order_votes[commit_id]| >= 2f+1 {
+	    // Add order_vote_msg.qc to the block store, if not already present
+		// Send the block with commit_id to execution
+		// Update highest_ordered_cert
+		finalize(order_vote.ledger_info)
+	}
+}
+```
+
+### Proof of Safety
```
danielxiangzl, line 137: I think it is better to remove proofs in this AIP. They are not rigorous enough.

zekun000 (approved):

aips/aip-order-votes.md
vusirikala: @zekun000 I see you approved already. Is it okay if I remove the proofs?

zekun000: yeah I don't think it's necessary to have proofs here

vusirikala: Done.

thepomeranian (approved):

PR #454, opened 2024-06-14
Update aip-80.md

alinush:
Simple clarification that this can happen by confusing the SK with an address and copy-and-pasting it in the wallet when sending a payment to oneself.


alinush:
@gregnazario should we add this clarification, or no?

gregnazario (approved):

thepomeranian (approved):

PR #475, opened 2024-07-15
Collection Permissions Update

JohnChangUK:
## Description
AIP - Collection Permissions Update

This AIP goes over updating permissions over a collection to adjust to the collection `owner` rather than the `creator`.


bowenyang007 (approved):

bowenyang007:
uh i don't have write access lol. Who can approve this.

aips/aip-95.md
```diff
@@ -0,0 +1,195 @@
...
+
+- The data was conducted by querying all collections, excluding collections owned by the dead wallet: `0xffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff`.
+
+**TransferRef not stored for the collection**
+
+As collections are initially created with `ungated_transfer` as false, this means collections can only be transferred with a `TransferRef`.
+If this was not created during construction of the collection, then this collection cannot be transferred to a new owner, meaning no other address can mint tokens from the collection.
+
+To mitigate this, the creation of the collection will not have `ungated_transfer` disabled. This change is reflected above
+in `collection.move` under `Specification`.
```
davidiw, line 181: I'm not sure how I feel about this. I think it makes more sense to let this be configurable during collection generation, but I would prefer this remains disabled due to the security properties resulting from it being otherwise.

davidiw (commented):
also it might make sense to actually expose a `MintRef` even if we don't use it....

thepomeranian (approved):

PR #479, opened 2024-07-21
[AIP] Add draft AIP for consensus observer.

JoshLind:
This PR adds a draft AIP for consensus observer.

### DIFF
--- a/aips/aip-93.md
+++ b/aips/aip-93.md
@@ -1,0 +1,200 @@
+---
+aip: 93
+title: Consensus Observer
+author: Josh Lind (https://github.com/JoshLind) & Zekun Li (https://github.com/zekun000)
+discussions-to (*optional):
+Status: Draft
+last-call-end-date (*optional):
+type: Core
+created: 07/19/2024
+---
+# AIP-93 Consensus Observer
+
+## Summary
+
+Aptos nodes currently use [state synchronization](https://medium.com/aptoslabs/the-evolution-of-state-sync-the-path-to-100k-transactions-per-second-with-sub-second-latency-at-52e25a2c6f10)
+(state sync) to stay-up-to-date with the blockchain. When validators execute a new block of transactions, state sync
+will synchronize the block data to validator fullnodes (VFNs) and public fullnodes (PFNs), ensuring all nodes have the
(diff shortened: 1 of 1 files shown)


aips/aip-xx.md
```diff
@@ -0,0 +1,198 @@
...
+
+There are several risks and drawbacks associated with CO, including:
+1. **Transaction Re-execution**: One of the benefits of state sync is that it provides an
+   [intelligent syncing](https://aptos.dev/en/network/nodes/configure/state-sync#intelligent-syncing)
+   mode, where fullnodes can skip transaction re-execution and reuse the execution results from validators.
+   This helps to reduce CPU usage and improve synchronization time. However, CO does not
+   provide this as it requires fullnodes to re-execute blocks in real-time. As a result, it requires all fullnodes
+   to have equivalent (or better) hardware resources than validators, to ensure they can process blocks at the
+   same rates while dealing with other overheads (e.g., REST API requests).
+      - **Mitigation**: While this is a drawback, it does not pose a significant risk, as fullnodes are already
```
sherry-x, line 154: Can a fullnode choose to not use CO, accept the latency penalty at high load, and potentially save hardware cost if they want to? or is this going to be an enforced sync mode for all nodes?

aips/aip-xx.md
```diff
@@ -0,0 +1,198 @@
...
+      required to deploy equivalent hardware resources as validators. Moreover, the benefits of CO in terms of
+      reduced latencies are more than sufficient to outweigh the additional costs of transaction re-execution.
+2. **CO Complexity**: CO will introduce additional complexity to the Aptos node software, as new components
```
sherry-x, line 157: Is there any extra bandwidth cost because of the new network messages?

aips/aip-xx.md
JoshLind: Yes, it's config gated, so nodes can opt-out and use regular state sync if they want. I'll update the text 😄

aips/aip-xx.md
JoshLind: No, if anything it'll be less data because nodes will be syncing transactions instead of outputs. 😄

sherry-x (approved):

PR #111, opened 2023-04-24
[aip-21] update api name

lightmark:
- change `allow_ungated_balance_transfer` to `frozen`.
- change `set_ungated_transfer` to `set_frozen_flag`.
match https://github.com/aptos-labs/aptos-core/pull/7908

PR #481, opened 2024-07-29
[aip proposal] Enum Types in the Move VM

wrwg:
AIP for enum types (only VM).


aips/aip-xx-enums.md
```diff
@@ -0,0 +1,253 @@
...
+
+There are many advantages of folding enums into the existing struct representation. Specifically since backends consume serialized representations of structs, and don’t care whether a blob is an enum or not, the enum extension does not cascade into those components. Also, for RAM representation of enums this viewpoint is suitable. Like in a C union, the variant tag determines the interpretation of the memory blob, but otherwise it behaves like a struct.
+
+A number of new instructions need to be introduced to deal with enums, notably a new form of borrow operation on fields of variants that aborts for unexpected variants. The later one is needed for allowing frontends to generate matching code via references to the data.
+
+## Impact
+
+This is a significant extension to the Move VM which will highly improve the fidelity of the code which can be written in Move. This first version contains the minimal necessary set of new features, more are needed in the future for improving performance.
```
igor-aptos, line 77: what does 'fidelity' refer to here? 

I haven't heard that term being used for writing code before.

aips/aip-xx-enums.md
```diff
@@ -0,0 +1,253 @@
...
+
+## Alternative solutions
+
+There are no viable alternatives known. The closest would be to represent a variant type as a struct with optional fields, but this is inefficient and would not give upgradability. 
+
```
igor-aptos, line 82: should we add something like:

If we think about upgradeability problem itself, serialization protocol could be changed/extended, to support upgradable structs. But that a significant change to the serialization protocol, and enum's solve variety of other needs as well.

aips/aip-xx-enums.md
```diff
@@ -0,0 +1,253 @@
...
+### Upgrade Compatibility
+
+Variant structs can be upgraded by adding new variants. The new variants must be added, in sequence, after the existing ones which must be a prefix of the variants of the new variant struct.
+
+### Handles
+
+A number of new handle types are needed to access variant structs from instructions. The general principle of Move bytecode is that arguments to an instruction are represented as indices in tables of such handles.
+
+First a givem variant of a struct needs a denotation for operations like pack, unpack, and variant testing:
```
igor-aptos, line 130: givem => given

aips/aip-xx-enums.md
```diff
@@ -0,0 +1,253 @@
...
+The extension is neutral in serialization size.
+
+
+### Representation as VM 1.0 Value Types
+
+In the current Move VM value system, an enum is represented as a container of values (similar to structs and vectors). The first element in this container is a `u16` with the variant tag, the remaining values are interpreted depending on the variant.
+
+### Serialization
+
+The serialization format starts with the `u16` variant tag, followed by the values of the fields of that variant. During type-driven deserialization, first the variant tag is read, then the according fields with types derived from the chosen variant.
```
igor-aptos, line 120: is this compatible with how rust enums are serialized via bcs?

we do have places where rust needs to deserialize move struct, and we create an equivalent rust struct with appropriate annotations, and use bcs to deserialize it, relying on bcs doing the same thing.

(and if it is compatible with rust enum bcs serialization, we should call it out / have a unit test for it)

aips/aip-xx-enums.md
```diff
@@ -0,0 +1,253 @@
...
+}
+
+```
+
+Second, since field borrowing for variant structs is supported, a way to denote a field in a variant is required:
+
+```rust
+pub struct VariantFieldHandle {
+    pub owner: StructDefinitionIndex,
+    pub variants: Vec<VariantIndex>,
```
igor-aptos, line 145: VariantIndex is not defined here, would be useful to have.

does index in this vector mean anything, or is this just a "set" of viable variants? how does this handle if field is different for different variant? (i.e. if same name of the field has different positions in different variants)

if this is set would it be better to have it as HashSet or bit vector?

aips/aip-xx-enums.md
wrwg: Its commonly used in my world, but replace by 'quality' to avoid confusion

aips/aip-xx-enums.md
wrwg: Added the paragraph

aips/aip-xx-enums.md
wrwg: Now it is after latest PR. Also updated the text and called out the differences between BCS and JSON representation.

aips/aip-xx-enums.md
wrwg: Added definition of VariantIndex, and some more clarifying text

wrwg:
Updated, thanks and PTAL!

aips/aip-xx-enums.md
wrwg: Done

igor-aptos (approved):

thepomeranian (approved):

Niftonlevrai:
Hello, I don't know who to contact, so I'm trying my luck here,
I've been trying to vote since this morning for the Aptos Proposal: #111
Enable AIP 91 - compatibility checker v2 and enum types; but, when after entering the apt amount and clicking on YES, an error message appears: Voting temporary unavailable for your Pontem wallet. What should I do please

PR #492, opened 2024-08-29
[draft] AIP-X Disable safe burn of user owned objects

gregnazario:
This AIP proposes to disable safe burn of user owned objects.

### DIFF
--- a/aips/aip-x-disable-object-burn.md
+++ b/aips/aip-x-disable-object-burn.md
@@ -1,0 +1,139 @@
+---
+aip: (this is determined by the AIP Manager, leave it empty when drafting)
+title: Disable Object Burn
+author: gregnazario (greg@aptoslabs.com)
+discussions-to (*optional): <a url pointing to the official discussion thread>
+Status: Draft
+last-call-end-date (*optional): <mm/dd/yyyy the last date to leave feedbacks and reviews>
+type: Standard (Framework)
+created: 8/28/2024
+updated (*optional):
+requires (*optional):
+---
+
+# AIP-X - Disable Safe Burn of User-Owned Objects
+
+## Summary
+
+Safe burn was originally described in [AIP-45](./aip-45.md), which allowed users to tombstone their objects and remove
+direct linkage of an undeletable, also known as *soulbound*, object to an account. This AIP seeks to disable safe
+object burn, as it caused extra complexity, and sometimes unexpected consequences. Owners of burned objects will still
+be able to unburn and recover their tombstoned objects.
+
+### Goals
+
+As a result of this AIP, users will still be able to unburn their burnt objects, but will not be able to burn any new
+undeletable objects. This will reduce complexity and some interesting edge cases associated.
+
+If we delay enacting this, there may be more burnt objects, which could cause more unexpected behaviors around moving
+burnt objects to the burn address.
+
+### Out of Scope
+
+This removes the previous mitigation, since users cannot opt-out of receiving unsolicited content. Wallets or other
+platforms will need to come up with a standardized way of hiding unwanted items.
+
+Additionally, we will not be preventing unburning existing burnt objects at this time.
+
+## Motivation
+
+The purpose here is to remove burn to prevent confusion, and complexity around *soulbound* objects changing owners.
+Originally, this was meant mostly for being able to hide unwanted *soulbound* NFTs. But, there are more use cases for
+*soulbound* objects (e.g. primary fungible store), which can complicate functionality around them. Removing this
+functionality simplifies cases for builders, and can make it easier to accomplish things like allowlists on fungible
+assets.
+
+If we do not accept this proposal, some functionality around *soulbound* objects will require extra complexity or even
+may not be able to work around burns.
+
+## Impact
+
+This impacts the following parties:
+
+* Users who have already burnt, or plan to burn owned *soulbound* objects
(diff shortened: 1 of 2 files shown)


aips/aip-x-disable-object-burn.md
```diff
@@ -0,0 +1,139 @@
...
+assets.
+
+If we do not accept this proposal, some functionality around *soulbound* objects will require extra complexity or even
+may not be able to work around burns.
+
+## Impact
+
+This impacts the following parties:
+
+* Users who have already burnt, or plan to burn owned *soulbound* objects
```
sherry-x, line 53: can we show some data on how many burns happened in the past to understand the impact better?

aips/aip-x-disable-object-burn.md
```diff
@@ -0,0 +1,139 @@
...
+burnt objects to the burn address.
+
+### Out of Scope
+
+This removes the previous mitigation, since users cannot opt-out of receiving unsolicited content. Wallets or other
+platforms will need to come up with a standardized way of hiding unwanted items.
+
+Additionally, we will not be preventing unburning existing burnt objects at this time.
+
+## Motivation
```
sherry-x, line 38: The motivation here is not so clear to me. It mostly mentioned the complexity introduced, but is there a specific example to help understand this complexity better?

gregnazario: This leaves room for multiple vulnerabilities, that I did not want to mention.

aips/aip-x-disable-object-burn.md
gregnazario: I do not have this data as it is not indexed.

The complexity of handling indexing for something that was purely driven because a couple engineers didn't like having soulbound NFTs in their accounts is low.

It's not handled by any exchange, or by any NFT tooling, and is barely documented.

aips/aip-x-disable-object-burn.md
```diff
@@ -0,0 +1,139 @@
...
+
+Safe burn was originally described in [AIP-45](./aip-45.md), which allowed users to tombstone their objects and remove
+direct linkage of an undeletable, also known as *soulbound*, object to an account. This AIP seeks to disable safe
+object burn, as it caused extra complexity, and sometimes unexpected consequences. Owners of burned objects will still
+be able to unburn and recover their tombstoned objects.
+
+### Goals
+
+As a result of this AIP, users will still be able to unburn their burnt objects, but will not be able to burn any new
+undeletable objects. This will reduce complexity and some interesting edge cases associated.
```
davidiw, line 26:
```suggestion
As a result of this AIP, users will still be able to unburn their burnt objects, but will not be able to burn any new objects. This will reduce complexity and some interesting edge cases associated.
```

aips/aip-x-disable-object-burn.md
```diff
@@ -0,0 +1,139 @@
...
+
+If we delay enacting this, there may be more burnt objects, which could cause more unexpected behaviors around moving
```
davidiw, line 28:
```suggestion
If we delay enacting this, there may be more burnt objects, which increase risk to new protocols due to the nuanced behaviors around burnt objects.
```

aips/aip-x-disable-object-burn.md
```diff
@@ -0,0 +1,139 @@
...
+burnt objects to the burn address.
```
davidiw, line 29:
```suggestion
```

aips/aip-x-disable-object-burn.md
```diff
@@ -0,0 +1,139 @@
...
+---
+
+# AIP-X - Disable Safe Burn of User-Owned Objects
+
+## Summary
+
+Safe burn was originally described in [AIP-45](./aip-45.md), which allowed users to tombstone their objects and remove
+direct linkage of an undeletable, also known as *soulbound*, object to an account. This AIP seeks to disable safe
+object burn, as it caused extra complexity, and sometimes unexpected consequences. Owners of burned objects will still
+be able to unburn and recover their tombstoned objects.
```
davidiw, line 21: especially with the desire to create untransferable objects and the notion of ownership which becomes muddy when the burn address manages the object but the object is still owned by another address.

aips/aip-x-disable-object-burn.md
davidiw: how does it impact those that have already burnt?

davidiw (commented):
seems fine after a couple tweaks

aips/aip-x-disable-object-burn.md
```diff
@@ -0,0 +1,139 @@
...
+
+## Timeline
+
+### Suggested implementation timeline
+
+The expectation is that this should ship as part of the next release after being approved.
+
+### Suggested developer platform support timeline
+
+One day of work, to remove `burn` from any documentation on the [aptos.dev](https://aptos.dev) website.
```
davidiw, line 131: I think it should just be explicit callout that burn needs to be stricken from aptos.dev not an estimate on the amount of work

davidiw (commented):
seems fine after a couple tweaks

aips/aip-x-disable-object-burn.md
gregnazario: It does not impact already burnt objects.  However, objects that are burnt can still be unburnt.

Let me add that.

movekevin (approved):

thepomeranian (approved):

PR #510, opened 2024-09-27
Create Permissioned Signer AIP

runtian-zhou:
AIP for adding permissions to the signer.


aips/aip-x.md
```diff
@@ -0,0 +1,237 @@
...
+    }
+
+    public fun withdraw<T: key>(owner: &signer, store: Object<T>, amount: u64): FungibleAsset {
+        assert!(
+            permissioned_signer::check_permission(
+                owner,
+                amount as u256,
+                WithdrawPermission {
+                    metadata_address: object::object_address(&borrow_store_resource(&store).metadata)
+                }
```
davidiw, line 158: Maybe make `WithdrawPermission` an enum and place amount in the enum?
Do the casts to u256 cost extra?

aips/aip-x.md
```diff
@@ -0,0 +1,237 @@
...
+            ),
+            error::permission_denied(EWITHDRAW_PERMISSION_DENIED)
+        );
+        //… regular withdraw logic
+    }
+    
+    public fun grant_permission(
+        master: &signer,
+        permissioned: &signer,
```
davidiw, line 167: why isn't this the `Object<T>` permissioned?

aips/aip-x.md
```diff
@@ -0,0 +1,237 @@
...
+- Implemented the core library that generates and manages the permissioned signer.
+    - including Move and Rust components.
+- Sweeped throught Aptos Framework codebase with following permissions implemented:
+    - how much fungible asset/coin can be withdrawed from each permissioned signer.
+    - transfer Object.
+    - manipulate NFT object/collection.
+    - operations on account.
+    - allowed to vote
+    - allowed to stake
+    - Privileged operation where master signer is required.
```
davidiw, line 201: perhaps you could expand the AIP with snippets on how you handle all these operations? we should have the conversation on why these are the right way to do things by comparison to alternatives.

aips/aip-x.md
```diff
@@ -0,0 +1,237 @@
+---
+aip: (this is determined by the AIP Manager, leave it empty when drafting)
+title: Permissioned Signer
+author: runtian-zhou
```
davidiw, line 4: Perhaps we could be inclusive on authorship, I think minimally beyond yourself of course, we could add Wolfgang, Igor, myself? Maybe Aaron?

davidiw (approved):
Let's get more here, let's get comparison points. Though we can land and iterate.

Please get an issue added, so we can have appropriate side conversation.

aips/aip-x.md
```diff
@@ -0,0 +1,237 @@
...
+    - Work with our existing signer module so that users can still set constraints on what transactions invoking existing on chain move functions could invoke.
+
+
+## Impact
+
+This would be a full compatible change to the framework. Existing on chain move modules wouldn't be impacted. We will add those constraints to the Aptos Framework around popular asset types, i.e: Coin/FungibleAsset/Object/NFT.
+
+## Alternative Solutions
+
+Another way to enforce permission is to leverage AIP-56 the runtime access control engine in Move. However, the runtime access control is more suitable to provide coarse grained permissions around modules but it's hard to represent permissions such as withdrawing X APT from a signer. However, we are open to encoperate the changes to the Move on Aptos language extensions in the future if we want to extend the permission system to the general public.
```
igor-aptos, line 52: Last sentence seems unrelated to the first two, maybe should be a separate paragraph?

aips/aip-x.md
```diff
@@ -0,0 +1,237 @@
...
+
+    public fun signer_from_permissioned(p: &PermissionedHandle): signer;
+
+    public fun destroy_permissioned_handle(p: PermissionedHandle);
+
+    public fun is_permissioned_signer(s: &signer): bool;
+}
+```
+
+We would introduce a new struct called `PermissionedHandle` that is derived from the original signer. The handle is generated by `create_permissioned_handle_with_expiration`  API, which takes an original signer. In that handle, we will generate a fresh address using auid API to store the permission information that would be needed by the management API.
```
igor-aptos, line 81: code has create_permissioned_handle, while you use create_permissioned_handle_with_expiration here 

we should probably have `create_permissioned_handle` which only creates signer within the current transaction, and create_permissioned_handle_with_expiration which creates it for longer (but upper bounded, and by default short) time window

We need to make sure that default mode - which creates signer valid only within the current transaction, doesn't leak to other transactions in a block, which have same onchain timestamp (as timestamp gets changed only once per block)

aips/aip-x.md
```diff
@@ -0,0 +1,237 @@
...
+
+Once we have the handle, we can then invoke `signer_from_permissioned`  to obtain a permissioned signer value. The invariant would be that the address of the permissioned signer should be exactly the same as the existing master signer that is used to derive the handle. With the expiration date, this API would abort if the wall clock time has exceeded the expiration date. Also the handle should not be storable. The storable variant will need to follow the logic of account delegation. The handle is also not droppable, so that we can enforce the destruction logic by using `destroy_permissioned_handle`  API.
+
+Note that this handle does not have store capability. There will be storable variant of this handle as well but that would contain an expiry time and the creation of this handle can only be invoked in the Aptos Framework as this is equivalent to signer delegation and thus need special security measure.
```
igor-aptos, line 85: hm here we are saying that expiration is a separate concern, but above we treat it together.

probably worth rephrasing - `create_permissioned_handle` only creates a handle within the current transaction, and is not storable.

and then separate section for storable handle with expiration.

aips/aip-x.md
```diff
@@ -0,0 +1,237 @@
...
+enum Signer {
+    Master(AccountAddress),
+    Permissioned {
+        master_addr: address,
+        permission_addr: address,
+    }
+}
+```
+
+Note that we did change the serialization rules for signer which would be a non-backward compatible change. This should be fine in general as signer should not be storable. The only problem is we exposes `size_of_val`  API in our framework. Since we changed the serialization, this would break the existing semantics for this API. To mitigate this, we will manually make sure that  `size_of_val`  would calculate the size of signer to be a single account address instead of an address plus a variant.
```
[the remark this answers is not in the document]
igor-aptos: maybe size_of_val / to_bytes just should abort if signer is inside? (same as we do with delayed fields)

and then let relevant parties know?

because any other solution here is a bit suboptimal

igor-aptos (approved):
added some comments.

let's land and iterate, as things evolve

aips/aip-x.md
```diff
@@ -0,0 +1,237 @@
...
+
+### Lifecycle of a permissioned handle
+
+The lifecycle of the new signer will look like the following:
+
+```jsx
+module aptos_framework::permissioned_signer {
+    struct PermissionedHandle {
+        master_addr: address,
+        permission_addr: address,
```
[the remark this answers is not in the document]
runtian-zhou: I was trying to avoid `Object<T>` because then we need to have ObjectCore and those management resources stored somewhere.

aips/aip-x.md
runtian-zhou: The amount has to be in the value storage as this needs to be updated upon withdraw. We do plan to make a couple of items there to be enum tho once the enum is stablized.

aips/aip-x.md
runtian-zhou: Added!

aips/aip-x.md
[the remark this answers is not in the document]
runtian-zhou: I do think we should make it abort. With that said the logic needs to be gated with feature flag.

runtian-zhou (commented):
Merge it for now and we will keep working on this AIP with the community review

thepomeranian (approved):

PR #511, opened 2024-10-01
Create aip for Aptos Intent Framework

runtian-zhou:


aips/aip-x.md
```diff
@@ -0,0 +1,192 @@
...
+We see intent as a programmable lock for on chain assets. Meaning:
+- One party can declare:
+  - Type of Asset they would like to trade in
+  - Condition when this asset could be released
+- The framework provides mechanism to release the lock when such condition is met.
+
+Logic here can then be translated into a Move struct that looks like the following:
+```
+struct TradeIntent<Source, phantom Target, Args> has key {
+    offered_resource: Source,
```
ch4r10t33r, line 39: Does `dispensable_resource` sounds better here?

aips/aip-x.md
```diff
@@ -0,0 +1,192 @@
...
+    consumption_function: FunctionInfo,
+    self_delete_ref: DeleteRef,
+    expiry_time: u64,
+}
+```
+
+The `TradeIntent` is parameterized by three type parameters:
+1. `Source`: Value that intent issuer is willing to give away
+2. `Target`: the type of value that the intent issuer is expecting.
+3. `Argument`: Intent user will use `Argument` as auxillary info to determine whether the unlock condition is met.
```
ch4r10t33r, line 51: Do you intend for `argument` to be a proof (zk-proof or storage proof), which can be verified on-chain?

aips/aip-x.md
```diff
@@ -0,0 +1,192 @@
...
+
+The lifecycle of an intent will look like the following:
+1. User sends a transaction to the blockchain to create the trading intent by specifying the resource they can offer and the expected condition when such resource can be unlocked.
+2. The intent would be broadcasted using an event.
+3. Anyone can execute this intent as long as the `consumption_function` can be executed successfully.
+    - We will expect intent solvers to actively monitor the available intents on chain and create transactions to claim the intent.
+
+With those types being generic, we can implement different trading intents easily:
+- If `Target` is instantiated with `FungibleAsset`, it means we are trading a Move value for a specific fungible asset.
+- If `Target` is instantiated with `LinearTransferRef`, it means we are trading a Move value for an Object.
```
ch4r10t33r, line 63: Should we include `NonFungibleAsset` as a type as well? Certain DeFi protocols can mint an NFT for a certain staking/deposit operation, no?

runtian-zhou: Here are just examples. You can put arbitrary type here.

aips/aip-x.md
runtian-zhou: In the current design, there's no proof involved: user specifies the post condition the transaction will need to meet, and aptos framework will assert this transaction can only be committed if post condition has been met. `Argument` is just some auxillary data to help the framework determine whether the transaction can be committed. It would be easier if you look at the move code directly.

aips/aip-x.md
```diff
@@ -0,0 +1,247 @@
...
+created: 09/27/2024
+updated (*optional): <mm/dd/yyyy>
+requires (*optional): <AIP number(s)>
+---
+
+# AIP-X - Aptos Intent Framework
+
+## Summary
+
+Designing a generic mechanism for declaring intents for trading between on chain resources on Aptos. This would allow users to declare which operations they would like to perform upfront and let 3rd party builders to fill in the details on how this transactions should be executed.
```
igor-aptos, line 18: this is not necessarily for trading, but for creating offers. (i.e. offer can be more general than a trade)

For example, you could have an offer for a bet.

aips/aip-x.md
```diff
@@ -0,0 +1,247 @@
...
+
+### Out of scope
+
+We will only be covering the intent declaration and execution logic in the framework. We are not going to touch the solving logic on how intents should be fulfilled and we will leave it for builders on Aptos to figure out.
```
igor-aptos, line 22: in the framework => in this AIP

aips/aip-x.md
```diff
@@ -0,0 +1,247 @@
...
+Intents rely on the existence of "solvers", entities whose job it is to provide solutions to intents. Solvers help find solutions to user-desired outcomes, and can get paid for doing so.
+
+This has a couple key benefits for users and developers:
+
+- It pushes sophistication and cognitive overhead from users / clients to more specialized entities
+- It allows other entities to compete to provide the best solution for a user, often resulting in better execution / prices
+- It can simplify and speed up cross-chain interactions by pushing risk to "relayers" or "fillers" in exchange for a small fee.
+- Intents are inherently secure, since the outcome signed by the user must match the outcome of the solution. However, ensuring that this is possible typically requires an "intent settlement network".
+
+Intent is a particular concept for Aptos because Move has a unique way of encoding on chain assets. With Move, you can represent
```
igor-aptos, line 42: sentence not finished?

aips/aip-x.md
```diff
@@ -0,0 +1,247 @@
...
+2. Intent needs to be broadcasted to the interested solver.
+    - In our initial implementation, intent creation could be done by executing a transaction. Such transaction can then emit an event to an event stream that solvers can subscribe themselves to.
+3. Intent Solver will maintain a local pool of intents for each type of intent contract and sythesize a path that would meet the exchange contract.
+4. Intent Solver will sign and submit transactions to process the intent.
+    - In our initial implementation, intent resolution is implemented first come first serve based. The first transaction that can resolve the intent will be able to get the resources wrapped in the intent.
+    - Intent resolution will be done by aptos framework code to make sure the solver did meet the condition listed in the intent contract.
+
+Aptos Labs will be responsible for implementing:
+1. A set of intent contracts to ramp up the system.
+2. Framework code for creating and validating intents.
```
igor-aptos, line 65: Framework code for creating and validating intents.
=> 
Library code for creating and validating intents.

aips/aip-x.md
```diff
@@ -0,0 +1,247 @@
...
+2. The intent would be broadcasted using an event.
+3. Anyone can execute this intent as long as the `Witness` can be provided.
+    - We will expect intent solvers to actively monitor the available intents on chain and create transactions to claim the intent.
+
+With those types being generic, we can implement different trading intents easily:
+- If `Target` is instantiated with `FungibleAsset`, it means we are trading a Move value for a specific fungible asset.
+- If `Target` is instantiated with `LinearTransferRef`, it means we are trading a Move value for an Object.
+- If `Source` is instantiated with `TransferRef`, it means we are giving out ownership to an Object given certain condition.
+- If `Source` is instantiated with `FungibleAsset`, it means we are giving out a certain fungible asset given certain condition.
+   - This doesn't quite work because `FungibleAsset` is not storable. The real implementation would look like giving out ownership of a `FungibleStore`
```
igor-aptos, line 133: could this be hidden in the intent? i.e. it stores FungibleStore, but returns extracted FungibleAsset to the caller?

aips/aip-x.md
```diff
@@ -0,0 +1,247 @@
...
+```
+public fun create_intent<Source: store, Args: store + drop, Witness: drop>(
+    dispensable_resource: Source,
+    argument: Args,
+    expiry_time: u64,
+    issuer: address,
+    _witness: Witness,
+): Object<TradeIntent<Source, Args>> 
+```
+User can use this api to register a trading intent to offer a value of `Source` type. The intent can only be executed if anyone can provide a value of `Witness` type. Intent always have an expiry time so that issuer can claim the resource back after the expiry time. Note that the return value of this function is wrapped in an `Object`. This is because we want anyone to be able to claim this intent by providing the address of this object.
```
igor-aptos, line 145: > Intent always have an expiry time so that issuer can claim the resource back after the expiry time.

Isn't it the opposite? issuer should always be able to claim the resource back, but after expiry - matching will be prevented?

aips/aip-x.md
```diff
@@ -0,0 +1,247 @@
...
+
+Another note is that this procedure is an on-chain procedure. Ideally we should make this a signed message instead so that users don't need to publish this intent to everyone.
```
igor-aptos, line 147: this sentence probably goes to future work

aips/aip-x.md
```diff
@@ -0,0 +1,247 @@
...
+            error::invalid_argument(EAMOUNT_NOT_MEET),
+        );
+
+        primary_fungible_store::deposit(argument.issuer, received_fa);
+        intent::finish_intent_session(session, FungibleAssetRecipientWitness {})
+    }
+}
+```
+Here we are defining an example of intent where the unlock condition is receiving a certain amount of a given fungible asset type. `Argument` type is instantiated to `FungibleAssetLimitOrder` and `Witness` type is instantated to `FungibleAssetRecipientWitness` in this example. In the arugment, we specify the amount and the type of FA we would like to receive. The check will then be implemented by first checking if the passed in `FungibleAsset` meets the requirement set in the `FungibleAssetLimitOrder` and deposit the FA to the issuer's primary fungible storage if the conditions are met. Once the checks are completed, we will provide the witness `FungibleAssetRecipientWitness` to complete this intent session.
+
```
igor-aptos, line 202: this needs an example of snippet of code of what the issuer and settler would need to do , for the above fungible_asset_intent_hooks.

aips/aip-x.md
```diff
@@ -0,0 +1,247 @@
...
+public fun finish_intent_session<Witness: drop, Args: store + drop>(
+    session: TradeSession<Args>,
+    _witness: Witness,
+)
+```
+There are two apis associated with claiming/consuming an on-chain intent. The first API will allow anyone to consume an intent object. Consumption of the intent object can release the resource locked in the intent temporarily. There will also be a hot potato type returned called `TradeSession`. Since this is a hot potato type, the transaction sender will have no choice but invoke the second `finish_intent_session` api which will require a value of `Witness` type. 
+
+### Example: Defining an intent to receive a certain fungible asset.
+```
+module aptos_framework::fungible_asset_intent_hooks {
```
igor-aptos, line 173: it's not aptos_framework , but a different adress

aips/aip-x.md
```diff
@@ -0,0 +1,247 @@
...
+
+## Testing 
+
+Implemented tests in the framework.
+
+## Risks and Drawbacks
+
+### Frontrun Risk
+
+The intent claiming transaction can be front-runned by malicious mempool operator. Since anyone is able to sign transaction to claim intent solution, a malicous mempool operator can take the intent solving transaction submitted by the solver and sign the same transaction with its own address to be able to claim the lucrative ones. However, since txns usually won't stay too long in the mempool right now, it should not be a major concern at this point. In the future we might need some authentication schema for intent claiming process.
```
igor-aptos, line 216: I would rephrase this:

> As the system is first-come first-serve, the intent claiming transaction can be front-runned by whoever sees it - mempool operator, validator operator, etc. Since anyone is able to sign transaction to claim intent solution, anyone can take the intent solving transaction submitted by the solver and sign the same transaction with its own address to be able to claim the lucrative ones. If main value comes from finding the route, and not being willing to settle, privacy of submitted transactions will become imporatnt. Intent Solver might want to run a validator node or partner with validator operator. Additionally in the future, support for "secret transactions" , that are revealed only at the time of execution might be added.

aips/aip-x.md
```diff
@@ -0,0 +1,247 @@
...
+## Security Considerations
+
+WIP
+
+## Future Potential
+
+- We need offchain declaration of the intents
+- We can use this standard to incentivize for a credit score system for different DeFi lending protocol:
+    - Implement borrow/lending action as an NFT, and implement scoring system for the NFT as the consumption condition.
+
```
igor-aptos, line 228: "secret transactions"

igor-aptos (changes requested):
You should add a note about this not being published at 0x1. 

you can look at https://github.com/aptos-foundation/AIPs/blob/main/aips/aip-72.md#risks-and-drawbacks for what they wrote there, but it could probably be explained a bit more (and the reasons for going that way)

aips/aip-x.md
runtian-zhou: Yes it is currently hidden for implementation.

runtian-zhou:
Update the PR with @igor-aptos's comments

PR #467, opened 2024-06-25
[AIP] Lite Account (Account v2)

lightmark:
Lite Account, aka Account V2 AIP.

PR #512, opened 2024-10-02
[AIP] Account Abstraction

lightmark:
AIP for account abstraction on Aptos

### DIFF
--- a/aips/account_abstraction.md
+++ b/aips/account_abstraction.md
@@ -1,0 +1,213 @@
+---
+aip: 
+title: Account Abstraction
+author: lightmark
+discussions-to (*optional): <a url pointing to the official discussion thread>
+Status: Draft
+last-call-end-date (*optional): <mm/dd/yyyy the last date to leave feedbacks and reviews>
+type: Framework
+created: 10/01/2024
+updated (*optional): <mm/dd/yyyy>
+requires (*optional): <AIP number(s)>
+---
+
+# AIP-X - Account Abstraction
+
+## Summary
+
+Account Abstraction (AA) on Aptos allows any account to be authenticated through move code in addition to existing 
+authentication schemes supported by native code. This AIP will introduce a new authenticator called `Abstraction` that
+delegates the authentication verification to AptosVM running move-based authentication logic.
+
+### Out of scope
+
+How to properly design an authentication function/module is out of scope.
+The `PayMaster`, aka, online gas payer, is not included in the design of AA in this AIP.
+
+## High-level Overview
+The core concept is to enable the move function to authenticate account signers, moving beyond the limitations of native Rust authenticators, which only support a restricted set of cryptographic schemes. To accomplish this, we propose transferring the authentication process from native Rust code to the AptosVM layer for each abstracted account. The solution we propose involves the following steps:
+
+1. Store the `FunctionInfo` of the customized Move authentication function in a new resource at the AA (Abstract Account) address.
+2. When the account is configured to support an `AbstractionAuthenticator`, the Move function stored in step 1 will be executed (leveraging native dynamic dispatch) to perform the authentication check.
+
+### Why is this approach superior to other options?
+
+- No major security concerns: Users who opt for AA are responsible for managing their account’s security.
+- Scalable: This approach can easily scale as it decouples authentication from rigid native cryptographic checks and supports multiple authentication functions.
+- Audit-proof and safe: The dispatchable function mechanism has already been audited and is widely used in the FA standard, ensuring its security.
+- Rich transaction context: The Transaction Context already provides comprehensive information about the current transaction, which can be leveraged for robust authentication.
+
+This solution enhances flexibility and scalability while maintaining a secure and efficient authentication process.
+
+## Impact
+
+Account Abstraction(AA) aims to support dispatchable authentication **at the level of smart contract.**
+The account can set up in such a way that the transaction can be authorized with complicated move smart contract instead of native account authenticators provided by Aptos blockchain core, enabling a series of possibilities:
+
+- Multisig v2 account
+- Sandbox or temporary account to isolate assets/risks, etc.
+- Restricted permissions granted to trusted applications so users don’t need to sign every transaction.
+- Delegation of asset management to a different account (semi-custodial model)
+
+## Alternative Solutions
+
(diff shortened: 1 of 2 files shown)


aips/account_abstraction.md
```diff
@@ -0,0 +1,204 @@
...
+
+## Impact
+
+Account Abstraction(AA) aims to support dispatchable authentication **at the level of smart contract.**
+The account can set up in such a way that the transaction can be authorized with complicated move smart contract instead of native account authenticators provided by Aptos blockchain core, enabling a series of possibilities:
+
+- Multisig v2 account
+- Sandbox or temporary account to isolate assets/risks, etc.
+- Restricted permissions granted to trusted applications so users don’t need to sign every transaction.
+- Delegation of asset management to a different account (semi-custodial model)
```
yeptos, line 50: it'd be nice if a bit more detail can be provided here for each item. i'm curious how this can be related to multisig v2 accounts

aips/account_abstraction.md
```diff
@@ -0,0 +1,204 @@
...
+This solution enhances flexibility and scalability while maintaining a secure and efficient authentication process.
+
+## Impact
+
+Account Abstraction(AA) aims to support dispatchable authentication **at the level of smart contract.**
+The account can set up in such a way that the transaction can be authorized with complicated move smart contract instead of native account authenticators provided by Aptos blockchain core, enabling a series of possibilities:
+
+- Multisig v2 account
+- Sandbox or temporary account to isolate assets/risks, etc.
+- Restricted permissions granted to trusted applications so users don’t need to sign every transaction.
```
ch4r10t33r, line 49: Would it be more appropriate to include a provision for resource locks here? I made a similar suggestion in the [LightAccount AIP](https://github.com/aptos-foundation/AIPs/pull/467#discussion_r1785581602), but I believe this feature could be integrated into this proposal as well.

aips/account_abstraction.md
lightmark: multisig v2 uses `MultisigAccount` as the account resource. 
And the sender of multisig v2 transaction is a normal sender. They could be an AA account. I think basically if you want to migrate an account from multisig v2 to AA account. We have to find a way first to convert multisig account to a normal account with AA auth to seamlessly migrate. I think you are asking about this.

But this^ seems unrelated to AA design itself. It would be worth another AIP to explain this migration option.

yeptos: Enabling AA for an account means registering an account resource that contains a dispatchable auth function, right?

If enabling AA can be done via say a private entry function call, I think there won't be any special treatment needed for multisig v2 accounts. Since multisig v2 accounts are just accounts that happen to have a MulsitigAccount resource, they can just call the AA enabling function like any other accounts.

aips/account_abstraction.md
```diff
@@ -0,0 +1,213 @@
...
+```
+
+Any authentication function must have the same interface as the following function:
+```rust
+/// The native function to dispatch customized move authentication function.
+public fun dispatchable_authenticate(account: signer, signing_data: SigningData): signer;
+```
+The first parameter, `signer`, represents the current account that needs authentication, and the function's return value is also a `signer`. Typically, this function returns the input signer if the authentication check is successful. However, in some cases, a permissioned signer may be returned to control the resources the transaction can access.
+`auth_data` is an enum that contains all the data required to authenticate the transaction via account abstraction. The initial version just has two fields, where `digest` is the sha3 digest of the corresponding transaction signing message and `authenticator` is the attached byte array which serves as the proof of the authentication verification. Usually it encodes the signature over the digest at least.
+Best Practice: To prevent replay attacks (both cross-chain), the `authenticator` must depend on both `chain_id` and `digest`.
```
igor-aptos, line 113: isn't chain_id already part of digest, because it is part of RawTransaction?

aips/account_abstraction.md
```diff
@@ -0,0 +1,213 @@
...
+The feature flag for this feature is `std::features::ACCOUNT_ABSTRACTION`.
+
+## Testing
+The above PR has API e2e test the functionality of Account abstraction with various cases.
+
+## Risks and Drawbacks
+
+- AA is hard to adopt w/o an auth function library. To make it easy for user to use, the framework should provide a bunch of commonly used 
+authentication modules/functions for users to use. Otherwise, it is really hard for beginner to implement customized
+authentication logic in move.
```
igor-aptos, line 186: is this in scope of this AIP?

aips/account_abstraction.md
```diff
@@ -0,0 +1,213 @@
...
+
+- The gas limit of the auth function is fixed so AA cannot support complicated move functions.
```
igor-aptos, line 188: maybe `cannot support complicated` => `cannot support too complicated`, as everything is relative :)

igor-aptos (approved):
address comments inline, and land

aips/account_abstraction.md
```diff
@@ -0,0 +1,213 @@
...
+
+fun authenticate(account: signer, func_info: FunctionInfo, auth_data: AbstractionAuthData): signer acquires DispatchableAuthenticator {
+    let func_infos = dispatchable_authenticator_internal(signer::address_of(&account));
+    assert!(simple_map::contains_key(func_infos, &func_info), error::not_found(EFUNCTION_INFO_EXISTENCE));
+    function_info::load_module_from_function(&func_info);
+    dispatchable_authenticate(account, signing_data, &func_info)
+}
+
+/// The native function to dispatch customized move authentication function.
+native fun dispatchable_authenticate(account: signer, signing_data: SigningData, function: &FunctionInfo): signer;
```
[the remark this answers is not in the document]
lightmark: yes. let me update.

aips/account_abstraction.md
lightmark: oh yes.

aips/account_abstraction.md
lightmark: I think it is more like a following work.

aips/account_abstraction.md
lightmark: I think they can coexist. Any problem for that. Once you get the master signer in any means, you can opt to remove any auth resources.

thepomeranian (approved):

PR #523, opened 2024-10-21
[AIP-x] stateless account update

lightmark:
stateless account aip for orderless txn.

### DIFF
--- a/aips/aip-x.md
+++ b/aips/aip-x.md
@@ -1,0 +1,102 @@
+---
+aip: (this is determined by the AIP Manager, leave it empty when drafting)
+title: Stateless Account Update
+author: lightmark, igor-aptos, davidiw
+Status: Draft
+type: Framework
+created: <10/21/2024>
+---
+
+# AIP-X - Stateless Account Update
+
+## Summary
+The Stateless Account update introduces an optimization to the existing Aptos account model, enabling gas savings by not requiring the explicit creation of an account resource unless necessary. In this update, accounts that do not have a pre-existing account resource are treated as valid with default behaviors, particularly defaulting the auth_key to the account’s address. The account resource will only be created when a specific action, such as rotating the authentication key, requires it.
+
+This update allows for more efficient resource management and reduced gas costs while maintaining backward compatibility with existing accounts and the Move VM.
+
+### Out of Scope
+The orderless transaction feature, which may eliminate the sequence number requirement in the future, is not part of this AIP but will be the prerequisite and complements the Stateless Account update.
+
+## High-level Overview
+Currently, the Aptos blockchain requires an explicit account resource for any account interacting with the system, which incurs gas costs. The Stateless Account update introduces a feature that allows accounts without a resource to be treated as existing with default behaviors, specifically:
+- The `auth_key` defaults to the account's address (auth_key = account_address).
+- Other fields in the account resource (e.g., sequence number) are unused unless explicitly needed.
+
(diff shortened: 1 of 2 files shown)


aips/aip-x.md
```diff
@@ -0,0 +1,103 @@
...
+- Other fields in the account resource (e.g., sequence number) are unused unless explicitly needed.
+
+The account resource will only be created when the user performs an action that overrides these default behaviors (e.g., rotating the auth_key), thus reducing the need for upfront resource creation.
+
+This update retains the current Account struct as is and optimizes the system to handle accounts without resources in a gas-efficient manner.
+
+Benefits:
+- Gas Savings: By avoiding the upfront creation of account resources, this update reduces gas costs for accounts that do not require immediate resource usage.
+- Efficiency: Stateless Accounts optimize resource usage for simple operations, minimizing the on-chain footprint for accounts that do not need complex functionality.
+- mBackward Compatibility: The existing account model is fully preserved, and this update works seamlessly with existing accounts and applications.
```
alnoki, line 32:
```suggestion
- Backward Compatibility: The existing account model is fully preserved, and this update works seamlessly with existing accounts and applications.
```

aips/aip-x.md
```diff
@@ -0,0 +1,103 @@
...
+
+
+Limitations:
+- No Advanced Features Without Resource Creation: Stateless Accounts cannot perform actions like key rotation or use capability offers until the account resource is created.
+- Default Behavior: Stateless Accounts will use default values, such as auth_key = account_address, until the account resource is created.
```
igor-aptos, line 37: based on below, this doesn't seem like a limitation. i.e. rotation will not be rejected, it's just that account will be created at that time, and so the gas would need to be charged. maybe reprhase a bit?

> - Advanced Features will require account creation: When Stateless Accounts perform advanced actions like key rotation or use capability offers, account resource will need to be created, and it's associated storage costs paid.

aips/aip-x.md
```diff
@@ -0,0 +1,103 @@
...
+The Move VM will handle these default behaviors in the prologue:
+
+```move
+if !exists<account::Account>(account_address) {
+    auth_key = account_address;
+}
+```
+
+### Sequence Number Considerations
+While the sequence number is currently required for transaction ordering, the sequence number is not created unless the account resource exists. The future introduction of orderless_txn will remove the sequence number requirement entirely, enabling fully stateless transactions.
```
igor-aptos, line 69: add that, call to get_sequence_number will return 0 for stateless accounts that only used orderless_txns thus far.

aips/aip-x.md
```diff
@@ -0,0 +1,103 @@
...
+### Migration Plan
+No migration is necessary for existing accounts. The Stateless Account feature primarily impacts new accounts or accounts that have not yet required an account resource. Existing accounts with resources will continue to function without any changes.
+
+## Risks and Drawbacks
+
+Gas Savings vs Functionality
+- While Stateless Accounts save gas by avoiding resource creation, users who require more advanced account features (such as key rotation) will still need to create an account resource, incurring gas costs at that time.
+
+Ecosystem Impact
+- No Capability Offers by Default: Stateless Accounts cannot issue capability offers until the account resource is created. Developers relying on this feature should ensure the resource is created when needed.
```
igor-aptos, line 86: should capability offer just create the account at that time?

igor-aptos (approved):

aips/aip-x.md
lightmark: yeah. removed.

thepomeranian (approved):

PR #524, opened 2024-10-26
AIP draft - hide unwanted soulbound objects, previously known as safe burn

gregnazario:
This functionality was removed previously due to issues around the transfers.  This should now no longer be the case since the transfer functionality was removed.


davidiw (approved):

davidiw:
@gregnazario can you submit an update with the latest AIP number and apply the settings to the AIP text as well?

PR #526, opened 2024-11-08
[AIP-X] - Enable interoperability for Federated Keyless Accounts for the same issuer

heliuchuan:


aips/aip-103.md
alinush, line 1: I think the AIP should explain how the ZK relation in AIP-61 will change. This is to ensure due diligence in reasoning about this change in the context of all other checks in the ZK relation.

aips/aip-103.md
```diff
@@ -0,0 +1,110 @@
...
+Risks
+- Developers need to not use `aud`-less accounts when not appropriate. This can be mitigated by the Aptos SDK default behavior so that developers must explicitly enable using `aud`-less accounts.
+- This introduces an additional proving path, one where the `aud` is not checked.  It is important that such proofs are rejected if the account requires `aud` to be present, as encoded in the KeylessPublicKey.
+- As circuit changes are needed to support `aud`-less accounts, a new ceremony will be needed to generate the proving key and verification key.
+- We want such accounts to be limited to Federated Keyless Accounts, as constructing Keyless Accounts without aud checks is unsafe.  This can be mitigated by the Aptos SDK disallowing `aud`-less accounts from being used as Keyless Accounts.  The prover will also reject proof requests for Keyless providers (as of now Google and Apple).  However, in a world where 3rd party provers are permitted, we cannot prevent developers from using `aud`-less accounts as Keyless Accounts, but developers would not have any incentive construct such accounts for their users (these accounts would be accessable by any other dApp, regardless of trust).
```
alinush, line 31: Typo: incentive **to** construct

aips/aip-103.md
```diff
@@ -0,0 +1,110 @@
...
+- The verification key will need an update, which will invalidate all existing proofs.  Additionally the prover will need to start proving with the new proving key right away after the update.  The prover has already been updated to support the proving key rotations and the SDK also supports state checks to invalidate old proofs.
+
+## Alternative solutions
+
+The alternative is to add an additional keyless public key type where the formula to compute the IdCommitment does not contain the `aud` at all.
+
+This is the advantage of explicit type safety as a completely new validation path would be implemented.  There would be no risk of such proofs being accepted for accounts that require `aud` to be present due to explicit differences in how the proof would be validated gated on the type of public key.
```
alinush, line 38: validated or gated?

aips/aip-103.md
```diff
@@ -0,0 +1,110 @@
...
+
+Thus if we can leverage the existing design, it would be preferable to do so.
+
+## Specification and Implementation Details
+
+This AIP's implementation has three parts -
+
+1. We add an additional private input into the prover.  This value will indicate whether the `aud` check is enabled.  
+- If it is enabled, the circuit will do the status quo set of verifications.  
+- If it is disabled, the circuit will use an empty `aud` value as a private input and skip checking the `aud` field.  A valid JWT still needs to be provided and the public inputs hash calculation will still implicitly include the `aud` field via the `IdCommitment` in the `KeylessPublicKey`, but it will be set to the empty value.  This means that proofs for `aud`-less accounts will be rejected if the account requires `aud` to be present, as the `IdCommitment` will be different.
```
alinush, line 53: Not sure what you mean by "use an empty `aud` value as a private input"?

"skip checking the `aud` field" --> phrased like this, it sounds super dangerous.

In general, the description here is poor/imprecise: there are several `aud` fields flying around, so just talking about the "`aud` field" in general is ambiguous: is it the one in the JWT? Is the one in the `IdCommitment`? is it the empty `aud`?

In addition to a more precise overview of how it works, we will need to actually show the modified ZK relation, highlighting the changed parts.

aips/aip-103.md
```diff
@@ -0,0 +1,110 @@
...
+
+## Alternative solutions
+
+The alternative is to add an additional keyless public key type where the formula to compute the IdCommitment does not contain the `aud` at all.
+
+This is the advantage of explicit type safety as a completely new validation path would be implemented.  There would be no risk of such proofs being accepted for accounts that require `aud` to be present due to explicit differences in how the proof would be validated gated on the type of public key.
+
+However the drawbacks include:
+- We need to add a new keyless public key type, which may not be needed if we can leverage the existing design.  And avoiding proliferation of keyless public key types is desirable.
+- Requiring imiplementation of a new authentication path in the authenticator, which may be error prone and takes additional engineering effort.
```
mstraka100, line 42: "imiplementation" typo

aips/aip-103.md
```diff
@@ -0,0 +1,110 @@
...
+- Requires more complex changes to the prover as it would need to support a different public inputs hash calculation in order to differentiate between accounts with and without `aud`.  Or it would need to use a different circuit versionentirely.
```
mstraka100, line 43: "versionentirely" typo

aips/aip-103.md
mstraka100, line 53: Small clarification here - the _circuit_ will skip the aud check, and will be given an empty string as the aud input _by the prover_.

aips/aip-103.md
```diff
@@ -0,0 +1,110 @@
...
+1. Write unit tests for the circuit to verify that it correctly handles the `aud` check.
+2. Write unit tests for the SDK to verify that it correctly instantiates accounts with and without `aud` checks.
+3. Do a manual end-to-end test in devnet/testnet via the SDK once the verification key is updated.
+4. Write smoke tests ensuring that `aud`-less accounts are rejected if the account requires `aud` to be present.
+
+## Security Considerations
+
+The core security considerations are:
+- Making sure the the circuit can securely support `aud`-less accounts.
+- Making sure that such proofs are rejected if the account requires `aud` to be present (as encoded in the `KeylessPublicKey`s `IdCommitment`).
```
mstraka100, line 70: I am not sure what this line means in saying "proofs are rejected". If the skip aud checks boolean is private, then the verifier will not have access to it. So we cannot accept or reject a proof based on its value

aips/aip-103.md
```diff
@@ -0,0 +1,110 @@
...
+
+Many customers of Auth0 and Cognito have applications with different branding within the same user-pool/tenant ecosystem.  Thus it is natural for such customers to use different application identifiers for their applications for organizational purposes.  This AIP will enable Aptos Federated Keyless Accounts to be interoperable across such applications.
+
+## Impact and risks
+
+- The relaxation of the `aud` field will allow for broader interoperability across applications within the same user-pool/tenant.  This allows for broader adoption of Aptos Federated Keyless Accounts in such user ecosystems.
+
+Risks
+- Developers need to not use `aud`-less accounts when not appropriate. This can be mitigated by the Aptos SDK default behavior so that developers must explicitly enable using `aud`-less accounts.
+- This introduces an additional proving path, one where the `aud` is not checked.  It is important that such proofs are rejected if the account requires `aud` to be present, as encoded in the KeylessPublicKey.
```
zjma, line 29: does it mean `aud_required` should be an onchain config put under the "admin account" of the dapp?
relying on user input is never safe...

zjma: by "admin account" i mean the jwk addr defined [here](https://github.com/aptos-foundation/AIPs/blob/main/aips/aip-96.md)...

aips/aip-103.md
```diff
@@ -0,0 +1,110 @@
...
+- The verification key will need an update, which will invalidate all existing proofs.  Additionally the prover will need to start proving with the new proving key right away after the update.  The prover has already been updated to support the proving key rotations and the SDK also supports state checks to invalidate old proofs.
+
+## Alternative solutions
+
+The alternative is to add an additional keyless public key type where the formula to compute the IdCommitment does not contain the `aud` at all.
+
+This is the advantage of explicit type safety as a completely new validation path would be implemented.  There would be no risk of such proofs being accepted for accounts that require `aud` to be present due to explicit differences in how the proof would be validated gated on the type of public key.
+
+However the drawbacks include:
+- We need to add a new keyless public key type, which may not be needed if we can leverage the existing design.  And avoiding proliferation of keyless public key types is desirable.
```
zjma, line 41: the proposed approach needs a flag in `KeylessPublicKey` to indicate "should skip aud", right?
that's unavoidably also a new keyless public key type?

heliuchuan: No. It would not.  skip aud would only be in the request to the prover and in the SDK to contruct a public key with aud being '' when calculating the IDC

aips/aip-103.md
heliuchuan: Well it wouldn't be user input, it would be the dapp developer, unless that's who you mean by user

aips/aip-103.md
heliuchuan: I mean the a proof for an audless account is not accepted by an account with the aud set.

aips/aip-103.md
zjma: it's eventually up to the user agent (which is under user's control) whether to go aud-less mode, no?
If the dapp developer wants ensure such invalid proofs are rejected, it can only be done on chain?

aips/aip-103.md
```diff
@@ -0,0 +1,170 @@
...
+
+Many customers of Auth0 and Cognito have applications with different branding within the same user-pool/tenant ecosystem.  Thus it is natural for such customers to use different application identifiers for their applications for organizational purposes.  This AIP will enable Aptos Federated Keyless Accounts to be interoperable across such applications.
+
+## Impact and risks
+
+- The relaxation of the `aud` field will allow for broader interoperability across applications within the same user-pool/tenant.  This allows for broader adoption of Aptos Federated Keyless Accounts in such user ecosystems.
+
+Risks
+- Developers need to not use `aud`-less accounts when not appropriate. This can be mitigated by the Aptos SDK default behavior so that developers must explicitly enable using `aud`-less accounts.
```
alinush, line 28: "Developers need to not use `aud`-less accounts when not appropriate" is not helpful.

You have to describe when it is and when it isn't appropriate.

aips/aip-103.md
```diff
@@ -0,0 +1,170 @@
...
+- This introduces an additional proving path, one where the `aud` is not checked.  It is important that such proofs are rejected if the account requires `aud` to be present, as encoded in the KeylessPublicKey.
```
alinush, line 29: Please state more precisely? I assume you are talking about the proving service:

> It is important that the prover service rejects audless proving requests for keyless accounts that require `aud` to be present and non-empty. This includes both normal keyless accounts (cite aip) and federated keyless accounts (cite aip).

I don't know how you can do such rejection though: the request does not include any information about the address (see [here](https://github.com/aptos-foundation/AIPs/blob/main/aips/aip-75.md#request-format))

aips/aip-103.md
alinush, line 1: Can you also describe the changes to the prover service's request format (see https://github.com/aptos-foundation/AIPs/blob/main/aips/aip-75.md#request-format)?

I presume there are no changes to the pepper service, and if so, you should also briefly mention this and maybe discuss why we don't need any changes.

aips/aip-103.md
```diff
@@ -0,0 +1,170 @@
...
+- As circuit changes are needed to support `aud`-less accounts, a new ceremony will be needed to generate the proving key and verification key.
+- We want such accounts to be limited to Federated Keyless Accounts, as constructing Keyless Accounts without aud checks is unsafe.  This can be mitigated by the Aptos SDK disallowing `aud`-less accounts from being used as Keyless Accounts.  The prover will also reject proof requests for Keyless providers (as of now Google and Apple).  However, in a world where 3rd party provers are permitted, we cannot prevent developers from using `aud`-less accounts as Keyless Accounts, but developers would not have any incentive to construct such accounts for their users (these accounts would be accessable by any other dApp, regardless of trust).
+- The verification key will need an update, which will invalidate all existing proofs.  Additionally the prover will need to start proving with the new proving key right away after the update.  The prover has already been updated to support the proving key rotations and the SDK also supports state checks to invalidate old proofs.
```
alinush, line 32: "invalidate all existing proofs" --> not very clear what proofs you are referring to.

I think you mean "invalidate all current proofs that are waiting to be sent to the chain by dapps"

aips/aip-103.md
```diff
@@ -0,0 +1,170 @@
...
+        \mathsf{exp\_date}, 
+        \mathsf{exp\_horizon}, 
+        \mathsf{iss\_val}, 
+        \mathsf{extra\_field}, 
+        \mathsf{header}, 
+        \mathsf{jwk}, 
+        \mathsf{override\_aud\_val}
+      ),\\
+      \textbf{w}_\mathsf{priv} = (
+        \mathsf{\textcolor{red}{skip\_aud\_check\_val}},
```
alinush, line 53: Can you please rename this `skip\_aud\_check`? No `_val` is needed here. This is just a boolean.

aips/aip-103.md
```diff
@@ -0,0 +1,170 @@
...
+1. Verify that the public inputs hash $\mathsf{pih}$ is correctly derived by hashing the inputs in $\textbf{w}\_\mathsf{pub}$ with $H\_\mathsf{zk}$ (as explained above).
+2. Check the OIDC provider ID in the JWT:
+   - Assert $\mathsf{iss\\_val}\stackrel{?}{=}\mathsf{jwt}[\texttt{"iss"}]$
+3. If using `email`-based IDs, ensure the email has been verified:
+   - If $\mathsf{uid\\_key}\stackrel{?}{=}\texttt{"email"}$, assert $\mathsf{jwt}[\texttt{"email\\_verified"}] \stackrel{?}{=} \texttt{"true"}$
+4. Check the user’s ID in the JWT:
+   - Assert $\mathsf{uid\\_val}\stackrel{?}{=}\mathsf{jwt}[\mathsf{uid\\_key}]$
+5. Check the address IDC uses the correct values:
+   - Assert $\mathsf{addr\\_idc} \stackrel{?}{=} H'(\mathsf{uid\\_key}, \mathsf{uid\\_val}, \mathsf{aud\\_val}; r)$
+6. If we are doing `aud` checks (i.e., $\mathsf{skip\\_aud\\_check\\_val} = \bot$)
```
alinush, line 77: Change the color of this bullet to highlight the changes, or put everything in bold? You can keep the old stuff uncolored/unbolded. This will avoid repeating it below like you are doing now.

aips/aip-103.md
```diff
@@ -0,0 +1,170 @@
...
+7. Check the EPK is committed in the JWT’s `nonce` field:
+   - Assert $\mathsf{jwt}[\texttt{"nonce"}] \stackrel{?}{=} H’(\mathsf{epk},\mathsf{exp\\_date};\rho)$
+8. Check the EPK expiration date is not too far off into the future:
+   - Assert $\mathsf{exp\\_date} < \mathsf{jwt}[\texttt{"iat"}] + \mathsf{exp\\_horizon}$
+9. Parse $\mathsf{extra\\_field}$ as $\mathsf{extra\\_field\\_key}$ and $\mathsf{extra\\_field\\_val}$ and assert $\mathsf{extra\\_field\\_val}\stackrel{?}{=}\mathsf{jwt}[\mathsf{extra\\_field\\_key}]$
+10. Verify the OIDC signature $\sigma_\mathsf{oidc}$ under $\mathsf{jwk}$ over the JWT $\mathsf{header}$ and payload $\mathsf{jwt}$.
+
+## Alternative solutions
+
+The alternative is to add an additional keyless public key type where the formula to compute the IdCommitment does not contain the `aud` at all.
```
alinush, line 96: `IdCommitment`

aips/aip-103.md
```diff
@@ -0,0 +1,170 @@
...
+- Requiring implementation of a new authentication path in the authenticator, which may be error prone and takes additional engineering effort.
+- Requires more complex changes to the prover as it would need to support a different public inputs hash calculation in order to differentiate between accounts with and without `aud`.  Or it would need to use a different circuit version entirely.
+
+Thus if we can leverage the existing design, it would be preferable to do so.
+
+## Specification and Implementation Details
+
+This AIP's implementation has three parts -
+
+1. We add an additional private input into the prover.  This value will indicate whether the `aud` check is enabled.
```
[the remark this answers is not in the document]
alinush: But you have to adjust the language, since here you are talking about a (non-existant) `dont_skip_aud_checks` input.

aips/aip-103.md
```diff
@@ -0,0 +1,170 @@
...
+
+## Specification and Implementation Details
+
+This AIP's implementation has three parts -
+
+1. We add an additional private input into the prover.  This value will indicate whether the `aud` check is enabled.  
+- If it is enabled, the circuit will do the status quo set of verifications.  
+- If it is disabled, the circuit will use an empty `aud` private input (as provided by the prover) and skip checking the `aud` field.  A valid JWT still needs to be provided and the public inputs hash calculation will still implicitly include the `aud` field via the `IdCommitment` in the `KeylessPublicKey`, but it will be set to the empty value.  This means that proofs for `aud`-less accounts will be rejected if the account requires `aud` to be present, as the `IdCommitment` will be different.
+
+2. The prover API will also require an update to allow for indiciating whether the `aud` check is enabled.  This will be done by adding a new boolean argument to the `prove` API.
```
alinush, line 115: As mentioned above, you should show the new API and highlight the changes. This will serve as a good reference for us too.

aips/aip-103.md
```diff
@@ -0,0 +1,170 @@
...
+
+Thus if we can leverage the existing design, it would be preferable to do so.
+
+## Specification and Implementation Details
+
+This AIP's implementation has three parts -
+
+1. We add an additional private input into the prover.  This value will indicate whether the `aud` check is enabled.  
+- If it is enabled, the circuit will do the status quo set of verifications.  
+- If it is disabled, the circuit will use an empty `aud` private input (as provided by the prover) and skip checking the `aud` field.  A valid JWT still needs to be provided and the public inputs hash calculation will still implicitly include the `aud` field via the `IdCommitment` in the `KeylessPublicKey`, but it will be set to the empty value.  This means that proofs for `aud`-less accounts will be rejected if the account requires `aud` to be present, as the `IdCommitment` will be different.
```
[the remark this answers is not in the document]
alinush: "the public inputs hash calculation will still implicitly include the `aud` field " --> Yes, but which `aud` field? Again, there are 3 flying around.

The same confusion later: "the account requires `aud` to be present"

I think you should rewrite this whole paragraph from scratch. Describe everything succinctly and clearly. Don't ever mention an unqualified `aud` field.

alinush (changes requested):
Thanks for including the ZK relation diff! 🙏 

See comments!

aips/aip-103.md
heliuchuan: Was talking about the validators

aips/aip-103.md
heliuchuan: Oh there are changes needed now thinking about it.  Added.

aips/aip-103.md
```diff
@@ -0,0 +1,206 @@
...
+## Summary
+
+This AIP proposes enabling Aptos Federated Keyless to be interoperable with dApps from the same issuer (user-pool/tenant).
+
+For IAM providers like Auth0 and Cognito, JWT tokens are scoped to a user-pool/tenant via the `iss` field, and they are also scoped to a specific application via the `aud` field.  This means that JWTs from the same issuer but with different `aud` values are from different applications and cannot be used to derive the same Aptos Federated Keyless Account even though they represent the same user identity within the same user-pool/tenant.
+
+Many customers of Auth0 and Cognito have applications with different branding within the same user-pool/tenant ecosystem.  Thus it is natural for such customers to use different application identifiers for their applications for organizational purposes.  This AIP will enable Aptos Federated Keyless Accounts to be interoperable across such applications.
+
+## Impact and risks
+
```
alinush, line 24: Did you mean to write "Impact" above this.

aips/aip-103.md
```diff
@@ -0,0 +1,206 @@
...
+- The relaxation of the `aud` field will allow for broader interoperability across applications within the same user-pool/tenant.  This allows for broader adoption of Aptos Federated Keyless Accounts in such user ecosystems.
+
+Risks
+- Developers need to not use `aud`-less accounts when they want their keyless accounts to scoped to their own application. This can be mitigated by the Aptos SDK default behavior so that developers must explicitly enable using `aud`-less accounts.
+- This introduces an additional proving path, one where the `aud` is not checked.  It is important that such proofs are rejected by the validator during transaction submission if the sending account requires `aud` to be present, as encoded in the KeylessPublicKey.
```
alinush, line 29: This phrasing makes it sound like there is an extra check that the validators must do, which is not the case: a validator cannot distinguish between a ZKP with `skip_aud_check = 1` and one with `skip_aud_check = 0`.

All the work is done in the ZK relation $\mathcal{R}$ itself which only skips the aud check if:
1. `skip_aud_check = 1`
2. the `aud` in the `IdCommitment`, passed in as a public input to $\mathcal{R}$, is empty

aips/aip-103.md
```diff
@@ -0,0 +1,206 @@
...
+## Alternative solutions
+
+The alternative is to add an additional keyless public key type where the formula to compute the `IdCommitment` does not contain the `aud` at all.
+
+This is the advantage of explicit type safety as a completely new validation path would be implemented.  There would be no risk of such proofs being accepted for accounts that require `aud` to be present due to explicit differences in how the proof would be gated on the type of public key.
+
+However the drawbacks include:
+- We need to add a new keyless public key type, which may not be needed if we can leverage the existing design.  And avoiding proliferation of keyless public key types is desirable.
+- Requiring implementation of a new authentication path in the authenticator, which may be error prone and takes additional engineering effort.
+- Requires more complex changes to the prover as it would need to support a different public inputs hash calculation in order to differentiate between accounts with and without `aud`.  Or it would need to use a different circuit version entirely.
```
alinush, line 103: > "Or it would need to use a different circuit version entirely."

I am not sure what point this is trying to make, since this AIP's proposed solution also uses a different circuit entirely: the circuit has an extra `skip_aud_check` private witness input.

alinush: I guess you are trying to suggest that the upgraded circuit proposed by this AIP maintains the same public input format? 

Whereas the alternative solution would require a new public inputs format, where `skip_aud_check` is public.

And it would require new validation paths in the Aptos VM: a new `MaybeAudlessKeylessSignature` with a ZKP that verifies against this new public inputs format.

Plus, we'd still need to keep the old circuit around for backwards compatibility. (But keeping two VKs on chain requires other changes too.)

So, in the end, are you trying to suggest that this alternative requires the chain to have two VKs and all the associated changes in our signature validation paths?

aips/aip-103.md
```diff
@@ -0,0 +1,206 @@
...
+
+Thus if we can leverage the existing design, it would be preferable to do so.
+
+## Specification and Implementation Details
+
+This AIP's implementation has three parts -
+
+1. We add an additional private input, `skip_aud_check`, into the prover.  This value will indicate whether the `aud` check is enabled.
```
alinush, line 111: Into the circuit, not the prover service. The discussion for the prover service follows later on.

aips/aip-103.md
```diff
@@ -0,0 +1,206 @@
...
+- Requires more complex changes to the prover as it would need to support a different public inputs hash calculation in order to differentiate between accounts with and without `aud`.  Or it would need to use a different circuit version entirely.
+
+Thus if we can leverage the existing design, it would be preferable to do so.
+
+## Specification and Implementation Details
+
+This AIP's implementation has three parts -
+
+1. We add an additional private input, `skip_aud_check`, into the prover.  This value will indicate whether the `aud` check is enabled.  
+- If it is disabled, the circuit will do the status quo set of verifications.
```
alinush, line 112: Indent these a bit more so they are rendered nicely under the numbered bullet point above them.

aips/aip-103.md
```diff
@@ -0,0 +1,206 @@
...
+
+This AIP's implementation has three parts -
+
+1. We add an additional private input, `skip_aud_check`, into the prover.  This value will indicate whether the `aud` check is enabled.  
+- If it is disabled, the circuit will do the status quo set of verifications.  
+- If it is enabled, the circuit will use an empty string for the `aud` private input (as provided by the prover), which will be used as the `aud` value committed in the IdCommitment.  The circuit will skip matching the value of the JWT's aud claim with the `aud` private input (which is the empty string).  
+
+   Since the `aud` value is committed to in the `IdCommitment`, a proof generated with `skip_aud_check` will be rejected by the validator if the account is not `aud`-less, as the `IdCommitment` will be different, resulting in proof verification failing due to the public inputs hash being incorrect.
+
+   Similarly, a proof generated without `skip_aud_check` will be rejected by the validator if the account *is* `aud`-less, as in this case as well the `IdCommitment` will be different, resulting in proof verification failing due to the public inputs hash being incorrect.
```
alinush, line 117: Same issue here. I would rephrase as:

> Similarly, a proof generated with**out** `skip_aud_check` will be rejected by the validator if the account *is* `aud`-less, since the ZK relation will not be able to match the empty `aud` in the `IdCommitment` with the non-empty `aud` in the JWT.

aips/aip-103.md
```diff
@@ -0,0 +1,206 @@
...
+      pub exp_horizon_secs: u64,
+      pub pepper: Pepper,
+      pub uid_key: String,
+      pub extra_field: Option<String>,
+      pub aud_override: Option<String>,
+      pub skip_aud_check: bool, // New argument
+   }
+   ```
+
+3. The pepper API will also require an update to allow for indiciating whether the `aud` check is enabled.  This is because for the pepper is derived using the `aud` value as one of the inputs, and for audless accounts the `aud` value needs to be the empty string in order for the account to be used across applications (which will differ by the value of the `aud` claim in the JWT).  Thus the pepper API needs to know whether the account is `aud`-less to construct the pepper appropriately.  This will be done by adding a new boolean argument `skip_aud_check` to the `fetch_pepper` API.
```
alinush, line 138: Typo: indicating

alinush: Typo: "for the pepper" -> "the pepper" ?

aips/aip-103.md
```diff
@@ -0,0 +1,206 @@
...
+   ```rust
+   #[derive(Debug, Deserialize, Serialize)]
+   pub struct PepperRequest {
+      pub jwt: String,
+      pub epk: EphemeralPublicKey,
+      pub exp_date_secs: u64,
+      pub epk_blinder: Vec<u8>,
+      pub uid_key: Option<String>,
+      pub derivation_path: Option<String>,
+      pub skip_aud_check: bool, // New argument
```
alinush, line 149: I think the pepper service needs to limit uses of this parameter.

Specifically, it should probably only be allowed to be `true` for OIDC providers that we know are audless.

alinush: Can you add this discussion somewhere, if it isn't there?

aips/aip-103.md
```diff
@@ -0,0 +1,206 @@
...
+   ```
+
+4. The SDK will also need to be updated to support instantiating of such accounts.  This will require adding a new boolean argument to the `KeylessAccount` constructor and constructing the `KeylessPublicKey` and `AccountAddress` appropriately.
+
+## Testing (Optional)
+
+1. Write unit tests for the circuit to verify that it correctly handles the `aud` check.
+2. Write unit tests for the SDK to verify that it correctly instantiates accounts with and without `aud` checks.
+3. Do a manual end-to-end test in devnet/testnet via the SDK once the verification key is updated.
+4. Write smoke tests ensuring that `aud`-less accounts are rejected if the account requires `aud` to be present.
```
alinush, line 160: 5. Make sure pepper service and prover service rejects `skip_aud_check = true` for non-audless OIDC providers?

alinush (approved):
Left some more feedback. Please fix it and then let's merge it. We can address other things after.

aips/aip-103.md
heliuchuan: Correct

aips/aip-103.md
heliuchuan: no, just artifact from template

thepomeranian (approved):

PR #556, opened 2025-01-29
AIP-X Sign in with Aptos

aptqpv:
### Description

Initial draft for Sign in with Aptos, a standard for secure and user-friendly way to authenticate to off-chain resources using Aptos blockchain addresses.


aips/aip-x.md
```diff
@@ -0,0 +1,707 @@
...
+
+1. **Lack of standardization:** Applications on Aptos lack a unified authentication message format, leading to inconsistencies that confuse users.
+2. **Unreadable messages:** Authentication messages are displayed in plaintext, making them difficult for users to comprehend.
+3. **Limited wallet support:** Without standardization, wallets cannot effectively support authentication UIs.
+4. **Security risks:** Malicious websites can deceive users into signing messages intended for legitimate dApps.
+5. **Unintuitive processes:** The traditional `connect` + `signMessage` workflow involves two separate steps, creating a clunky and unintuitive user experience.
+
+Addressing these challenges is key to unlocking the full potential of decentralized authentication and delivering a seamless, secure, and user-friendly experience.
+
+![Figure 1: Sign in with Aptos](./static/8.png)
```
alinush, line 44: A `phishing.com` should **NOT** even be allowed to ask the user for a SIWA for `dapp.xyz`.

This picture suggests that it does.

I see no benefit to doing this. Only problems. Can we change this please?

aips/aip-x.md
```diff
@@ -0,0 +1,707 @@
...
+## High-level Overview
+
+**Decentralized authentication** is a privacy-first solution that empowers individuals to safeguard their information, personalize their experiences, and maintain freedom from censorship.
+
+By authenticating through a wallet, users have full control over their authentication journey. They can decide _what to share_ and _how to log in,_ ensuring a more secure and customizable experience.
+
+For example, a user may choose to use one of the wallet experiences to authenticate into an application:
+
+- **Petra** – _Passwords/Biometric_: - Fully non-custodial. Requires no personal information, ensuring complete privacy and resistance to censorship.
+- **Aptos Connect** – _Social Providers_: Fully non-custodial. Shares limited personal information like name and email but may be subject to censorship by social providers.
```
alinush, line 30: Please clarify **with whom** it shares the personal information.

Like we discussed before, I find this confusing: the chain / dapp does not learn anything, actually. 

Of course, the wallet (i.e., Aptos Connect) sees the JWT and learns your OIDC identity. I wouldn't necesarily call that "personal information" though; it's too broad. It's more helpful to be specific and say what it is: "name, profile picture, email, OIDC `sub`"

aips/aip-x.md
```diff
@@ -0,0 +1,707 @@
...
+
+> 💡 **_Decentralized authentication transforms wallets into decentralized identities._**
```
alinush, line 32: What is this? Why is it here?

alinush: This is a broad, vague claim that does not mean anything, AFAICT.

aips/aip-x.md
```diff
@@ -0,0 +1,707 @@
...
+
+However, while this approach offers significant benefits, it also presents several challenges:
+
+1. **Lack of standardization:** Applications on Aptos lack a unified authentication message format, leading to inconsistencies that confuse users.
```
alinush, line 36: **Missed opportunity:** Refer the reader to your screenshot that beautifully showcases this. Showing is better than telling.

aips/aip-x.md
```diff
@@ -0,0 +1,707 @@
...
+2. **Unreadable messages:** Authentication messages are displayed in plaintext, making them difficult for users to comprehend.
+3. **Limited wallet support:** Without standardization, wallets cannot effectively support authentication UIs.
+4. **Security risks:** Malicious websites can deceive users into signing messages intended for legitimate dApps.
```
alinush, line 39: Isn't the `signMessage` flow domain-separated? Doesn't it include the requesting dapp's URL in the signed message? I guess it is optional?

aips/aip-x.md
```diff
@@ -0,0 +1,707 @@
...
+5. **Unintuitive processes:** The traditional `connect` + `signMessage` workflow involves two separate steps, creating a clunky and unintuitive user experience.
```
alinush, line 40: Explain what are these two steps in parantheses.

aips/aip-x.md
```diff
@@ -0,0 +1,707 @@
...
+
+Addressing these challenges is key to unlocking the full potential of decentralized authentication and delivering a seamless, secure, and user-friendly experience.
+
```
alinush, line 43: The image here appears with any context or explanation.

Maybe move it in its own section and explain what the image showcases.

aips/aip-x.md
alinush: Is it `connect` + `signMessage`? 

I guess I don't understand how the `connect` step disappears. I suppose with SIWA, when you click "Connect wallet" you'll just get the `signIn` request before the wallet is actually connected, thereby merging everything into one step?

If so, this may be worth depicting via pictures & explaining.

aips/aip-x.md
```diff
@@ -0,0 +1,707 @@
...
+### For Developers
+
+- **Streamlined Integration:** SIWA provides a standardized framework for authentication, removing the need for custom implementations. This reduces development overhead and ensures compatibility across applications.
+- **Enhanced Security:** By enabling wallets to validate and flag potentially malicious messages, SIWA minimizes the risk of phishing attacks and other security vulnerabilities.
+- **Future-Proof Standards:** Building on existing specifications like EIP-4361 and CAIP-122, SIWA ensures that developers work with a modern, robust authentication protocol aligned with industry best practices.
+
+### For Users
+
+- **Improved Experience:** Replacing the cumbersome `connect` + `signMessage` flow with a single-step `signIn` method simplifies the login process, making it more intuitive and user-friendly.
+- **Transparency and Trust:** The standardized message format ensures that users can easily understand what they are signing fostering trust.
```
alinush, line 63: Add comma before "fostering"

aips/aip-x.md
```diff
@@ -0,0 +1,707 @@
...
+| address        | string          |           | Blockchain address performing the signing.                                                                                            |
+| uri            | string          |           | URI referring to the resource that is the subject of the signing i.e. the subject of the claim.                                       |
+| version        | string          |           | Current version of the message.                                                                                                       |
+| statement      | string          |           | Human-readable ASCII assertion that the user will sign. It MUST NOT contain `\n`.                                                     |
+| nonce          | string          |           | Randomized token to prevent signature replay attacks.                                                                                 |
+| chainId        | string          |           | Identifier for the network locating the address listed separately above.                                                              |
```
alinush, line 91: _"Identifier for the network locating the address listed separately above"_
--> 
"Identifier for the network where the `address` above lives (e.g., for Aptos `mainnet`, the chain ID would be 1)"

aips/aip-x.md
```diff
@@ -0,0 +1,707 @@
...
+| requestId      | string          |           | System-specific identifier used to uniquely refer to the authentication request.                                                      |
+| resources      | string[]        |           | List of information or references to information the user wishes to have resolved as part of the authentication by the relying party. |
+
+### AptosSignInMessage
+
+Once the fields are provided from the application, the wallet will be responsible for constructing the SIWA message. If the mandatory fields are not provided, they must be populated by the wallet.
+
+| Name                  | Type            | Mandatory | Description                                                                                                                           |
+| --------------------- | --------------- | --------- | ------------------------------------------------------------------------------------------------------------------------------------- |
+| fields.domain         | string          | ✓         | `dnsauthority` that is requesting the signing                                                                                         |
```
alinush, line 104: From the text, it is not clear: is `AptosSignInMessage::fields.domain` set to `AptosSignInInput::domain`?

I am a bit confused as to why you are you re-stating the same fields here.

Is there a less redundant way to describe this? May need to meet to clarify.

aips/aip-x.md
```diff
@@ -0,0 +1,707 @@
...
+
+### AptosSignInOutput
+
+When the user has approved and signed the message, the wallet will respond with an `AptosSignInOutput` to the application. The application should then use this response to verify if the outputs are valid with the inputs.
+
+| Name              | Type          | Mandatory | Description                                                                                                                        |
+| ----------------- | ------------- | --------- | ---------------------------------------------------------------------------------------------------------------------------------- |
+| account.address   | string        | ✓         | The address of the user that is signing in.                                                                                        |
+| account.publicKey | PublicKey     | ✓         | The public key of the user that is signing in.                                                                                     |
+| message           | SignInMessage | ✓         | A SignInMessage object that contains information about the message that was signed, it's signature, and plain text representation. |
```
alinush, line 128: Is this `SignInMessage` or `AptosSignInMessage`?

aips/aip-x.md
```diff
@@ -0,0 +1,707 @@
...
+| signature         | Signature     | ✓         | The signature of the message.                                                                                                      |
```
alinush, line 129: If we already had a signature above over (some serialization of) `AptosSignInMessage`, then what is this signature?

aips/aip-x.md
```diff
@@ -0,0 +1,707 @@
...
+- ${resources[n]}
+```
+
+#### Example Messages
+
+_Minimal Example_
+
+```
+example.com wants you to sign in with your Aptos account:
+0x0000000000000000000000000000000000000000000000000000000000000001
```
alinush, line 164: I don't think `0x000...0001` is a good example here: it will never arise in practice.

aips/aip-x.md
```diff
@@ -0,0 +1,707 @@
...
+Expiration Time: 2024-01-01T00:00:00Z
+Not Before: 2023-01-01T00:00:00Z
+Request ID: req123
+Chain ID: mainnet
+Resources:
+- resource1
+- resource2
+```
+
+### Augment Backus-Naur Form (ABNF) Expression
```
alinush, line 192: Augment**ed**

aips/aip-x.md
```diff
@@ -0,0 +1,707 @@
...
+```
+example.com wants you to sign in with your Aptos account:
+0x0000000000000000000000000000000000000000000000000000000000000001
+
+URI: https://example.com
+Version: 1
+Chain ID: mainnet
+```
+
+_Detailed Example_
```
alinush, line 171: It would be very helpful if this AIP described when field like `nonce` should be set in the message. 

This section on examples could be a good place to do that: go over a few sample messages and detail an application where that message suffices. I am actually wondering where the "minimal example" above would actually arise.

aips/aip-x.md
```diff
@@ -0,0 +1,707 @@
...
+```
+
+### Domain Separating Signing Algorithm
+
+A plaintext construction of a SIWA message must be prefixed with a valid SIWA hash before signing to ensure domain separation across other signatures. The messages must conform to the following expression:
+
+```
+sha3_256(b"SIGN_IN_WITH_APTOS::") || <siwa_message>
+```
+
```
alinush, line 235: Maybe add the following clarification:

 > where `<siwa_message>` is a sign-in with Aptos message formatted as explained in ["Message Format"](#message-format) above.

aips/aip-x.md
```diff
@@ -0,0 +1,707 @@
...
+### Wallet Standard
+
+The [Wallet Standard](https://github.com/aptos-labs/wallet-standard) is a chain-agnostic set of interfaces and conventions that aim to improve how applications interact with injected wallets.
```
alinush, line 238: Are you explaining what the wallet standard is because you are about to propose a modification/extension of it?

If so, maybe clarify from the beginning:

"This AIP also modifies / extends the wallet standard, which is a chain-agnostic set of [...]"

aips/aip-x.md
```diff
@@ -0,0 +1,707 @@
...
+
+#### Suggested Features
+
+The features below will be added to the wallet as suggested features. Features will be optional to maintain compatibility with existing wallets.
```
alinush, line 242: *to the wallet **standard**

no?

aips/aip-x.md
```diff
@@ -0,0 +1,707 @@
...
+`aptos:signIn` method to establish a connection and to sign a _Sign in with Aptos_ authentication message.
+
+```ts
+// `input: AptosSignInInput` - inputs with fields to construct an authentication message
+signIn(input: AptosSignInInput): Promise<UserResponse<AptosSignInOutput>>
+```
+
+### Wallet Implementer
+
+To ensure the safety of our users, we want to take a conservative approach when allowing users to sign potentially harmful messages.
```
alinush, line 253: We can't be **both** "allowing users to sign potentially-harmful messages" and "take a conservative approach."

I propose we nix the first and only do the 2nd :)

aips/aip-x.md
```diff
@@ -0,0 +1,707 @@
...
+
+#### Bound Fields Verification
+
+The following fields MUST be verified to prevent phishing attacks: `domain` , `address` , `uri` , `version` , and `chainId` . If the fields do not match the signing request, errors must be displayed to the users detailing them out. It is suggested to add additional steps to the approval process, such as an opt-in checkmark field, that confirms the user’s intent if errors are present.
```
alinush, line 257: Saying that "fields MUST be verified" is not precise:

1. Who is doing the verification?
2. What exactly is this verification for each field?

aips/aip-x.md
```diff
@@ -0,0 +1,707 @@
...
+
+A wallet may also decide to prevent the approval of invalid signing requests.
```
alinush, line 259: This should be a **MUST**, not a _may_.

aips/aip-x.md
```diff
@@ -0,0 +1,707 @@
...
+
+## Reference Implementation
+
+### Interfaces
+
+#### @aptos-labs/siwa
+
+A new package named `@aptos-labs/siwa` will be created to house the utilities for the _Sign in with Aptos_ standard. These utilities will be responsible for creating messages, verifying messages, and deserializing public key and/or signatures according.
```
alinush, line 267: NPM package? What kind of package?

alinush: What language?

aips/aip-x.md
```diff
@@ -0,0 +1,707 @@
...
+
+Below are the primary functions that that will be provided as part of the `@aptos-labs/siwa` utilities:
+
+```ts
+/**
+ * Create a SignIn message text from the input following the ABNF format defined in the Sign in with Aptos
```
alinush, line 273: Pro-tip: In technical writing, you always want to refer to the same thing by one name. It can't be either of `SignMessage`, `AptosSignInMessage` or `SignIn` message. You have to pick one and use it consistently so it is drilled down into your readers head and so they know what it means.

aips/aip-x.md
```diff
@@ -0,0 +1,707 @@
...
+ * specifications.
+ *
+ * @param input The input to create the SignIn message text from.
+ *
+ * @returns The SignIn message text.
+ */
+function createSignInMessageText(input: AptosSignInMessage["fields"]): string;
+
+/**
+ * Verifies an input SignIn message against expected fields.
```
alinush, line 283: Same problem here.

aips/aip-x.md
```diff
@@ -0,0 +1,707 @@
...
+| fields.resources      | string[]        |           | List of information or references to information the user wishes to have resolved as part of the authentication by the relying party. |
+| plainText             | string          | ✓         | The constructed message using the standardized message format.                                                                        |
+| signature             | bytes           | ✓         | The signature of the message. This must conform to the domain separated signing algorithm.                                            |
+| type                  | string          | ✓         | The public key scheme used to sign the message (e.g. 'ed25519', 'single_key', 'multi_key', 'multi_ed25519', etc.)                     |
+
+### AptosSignInOutput
+
+When the user has approved and signed the message, the wallet will respond with an `AptosSignInOutput` to the application. The application should then use this response to verify if the outputs are valid with the inputs.
```
alinush, line 122: Unclear what this means: "verify if the outputs are valid with the inputs."

aips/aip-x.md
```diff
@@ -0,0 +1,707 @@
...
+  input: AptosSignInInput,
+  output: {
+    publicKey: PublicKey;
+    signature: Signature;
+    message: string;
+  }
+): VerificationResult<AptosSignInMessage["fields"]>;
+
+/**
+ * Generate a signing message using the Sign in with Aptos signing algorithm.
```
alinush, line 313: Is a "signing message" different than a "[Aptos]SignInMessage"?

If so, did the AIPs text properly introduce both & differentiate between them? I don't think "signing message" has been introduced.

aips/aip-x.md
```diff
@@ -0,0 +1,707 @@
...
+ * sha3_256( sha3_256(b"SIGN_IN_WITH_APTOS::" ) || <message> )
+ *
+ * @param message The SIWA message to sign.
```
alinush, line 316: Again, what is "SIWA message"? It looks like another unnecessary synonym.

aips/aip-x.md
```diff
@@ -0,0 +1,707 @@
...
+ 
+ * @returns The signing message.
+ */
+function generateSignInSigningMessage(message: string): Uint8Array;
+```
+
+#### @aptos-labs/wallet-standard
+
+The following interfaces would be introduced into the `@aptos-labs/wallet-standard` package.
```
alinush, line 325: Include a link to it here, please.

aips/aip-x.md
```diff
@@ -0,0 +1,707 @@
...
+  nonce?: string;
+  chainId?: string;
+  issuedAt?: string;
+  expirationTime?: string;
+  notBefore?: string;
+  requestId?: string;
+  resources?: string[];
+};
+
+export type AptosSignInMessage = {
```
alinush, line 345: I think you should solely use this `AptosSignInMessage` term throughout the text, instead of synonyms for it.

aips/aip-x.md
```diff
@@ -0,0 +1,707 @@
...
+export type AptosSignInOutput = {
+  account: AccountInfo;
+  message: AptosSignInMessage;
+  signature: Signature;
+};
+```
+
+### Wallet Implementation
+
+The wallet is responsible for taking `AptosSignInInput` fields defined by an an application and returning a response object with a signature, plain text representation of the message, and account information.
```
alinush, line 374: Typo: "an an" --> an

alinush: What is a "response object"? Again, no synonyms.

alinush: What is "the message"?

aips/aip-x.md
```diff
@@ -0,0 +1,707 @@
...
+
+The wallet must bind the `address`, `chainId` , `domain` , `uri` , and `version` fields if not present. If there is a mismatch between these fields and the ones defined by the application, it is important that it is flagged and shown to the user.
```
alinush, line 376: Not sure I understand what "bind" means. You mean "set to the correct value"?

alinush: Again, "flagged" is not good enough. It must be forbidden.

aips/aip-x.md
```diff
@@ -0,0 +1,707 @@
...
+export class MyWalletAccount implements AptosWalletAccount {
+  async signIn(
+    input: AptosSignInInput
+  ): Promise<UserResponse<AptosSignInOutput>> {
+    // ...
+  }
+}
+```
+
+2. The Wallet will construct a `AptosSignInMessage` with the input. The `createSignInMessageText` utility from `@aptos-labs/siwa` can be used. Once the plaintext representation of the message has been constructed, we want to generate a signing message using `generateSignInSigningMessage` that conforms to _Sign in with Aptos_’s domain separated signing algorithm. **It is important to bind the following fields: `address`, `chainId` , `domain`, `uri` , and `version`.**
```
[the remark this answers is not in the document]
alinush: As mentioned before, not sure if **bind** is the right word here. If it is, it may be worth clarifying this earlier in the AIP.

aips/aip-x.md
```diff
@@ -0,0 +1,707 @@
...
+        signature,
+      },
+    };
+  }
+}
+```
+
+### Application Implementation
+
+The application will be responsible for sending the `signIn` request with `AptosSignInput` fields to the wallet. The application should typically create the fields in the backend and store them to verified at the end of the flow.
```
alinush, line 559: Typo: `AptosSignInput` --> `AptosSignInRequest`

aips/aip-x.md
alinush, line 1: A couple of high level thoughts:
- `AptosSignInInput` is not as descriptive of a name as `AptosSignInRequest`, no? It is a request, no?
- I propose you replace `AptosSignIn` with `SiWA` or `SIWA` everywhere. Not only less verbose, but also more to the point: this AIP is about SIWA.
- There is quite a bit of terminology here and the AIP currently uses too many names for the same term. This has to be fixed: one term, one name.
- You may even want to have a section where you overview the terminology. Not sure.
- I think the "Overview" should actually overview what is happening in SIWA: dapp makes a request to a wallet (in a certain format), wallet displays a prompt to the user, user clicks approve, wallet signs (in a certain format), dapp gets signature.

aips/aip-x.md
```diff
@@ -0,0 +1,707 @@
...
+};
+```
+
+### Wallet Implementation
+
+The wallet is responsible for taking `AptosSignInInput` fields defined by an an application and returning a response object with a signature, plain text representation of the message, and account information.
+
+The wallet must bind the `address`, `chainId` , `domain` , `uri` , and `version` fields if not present. If there is a mismatch between these fields and the ones defined by the application, it is important that it is flagged and shown to the user.
+
+#### Overview
```
alinush, line 378: I found this hard to read / understand. Not a TS guy so that may be why. But there may also be too many things here. Do you need all of this? Can anything be left out / simplified?

aips/aip-x.md
```diff
@@ -0,0 +1,707 @@
...
+    // ... the user is signed in, redirect to application
+  };
+
+  // ...
+}
+```
+
+## Testing
+
+The new `@aptos-labs/siwa` package will have its utilities tested to ensure that validation is done correctly. Coverage is most important here. In addition, signature verification methods will be used from the `@aptos-labs/ts-sdk`, it's important that public key verification methods are supported and tested.
```
alinush, line 670: > it's important that public key verification methods are supported and tested

What are you referring to here?

aips/aip-x.md
```diff
@@ -0,0 +1,707 @@
...
+
+## Risks and Drawbacks
+
+The ecosystem is responsible for adopting the new standard as an alternative to the `connect` + `signMessage` flow. If applications/wallets do not properly implement the `signIn` flow, they may suffer vulnerabilities that can lead to potential phishing, replay, or forgery attacks. If vulnerabilities are found within _Aptos Labs_ or ecosystem utilities, it’s important to patch and deprecate them as soon as possible. Applications that accept the `connect` + `signMessage` flow may still be vulnerable to security attacks.
```
alinush, line 674: > If applications/wallets do not properly implement the `signIn` flow, they may suffer vulnerabilities that can lead to potential phishing, replay, or forgery attacks.

I think the AIP actually needs to discuss ways in which applications can mess this up. Ideally, there should not be (m?)any.

alinush: > If vulnerabilities are found within Aptos Labs or ecosystem utilities, it’s important to patch and deprecate them as soon as possible.

This is self-evident. Not sure why you are stating it here.

aips/aip-x.md
```diff
@@ -0,0 +1,707 @@
...
+
+## Security Considerations
+
+It is important that signature verification methods are implemented correctly through the `@aptos-labs/ts-sdk`. If not correct, applications and wallets utilizing these methods will be vulnerable to attacks. Additional utilities provided from the `@aptos-labs/siwa` must be well tested to avoid attacks.
+
+The attack vectors that must be accounted for are the following:
+
+- Phishing attacks — A user must have visibility on signing a message that is for a different application.
```
alinush, line 682: Such a phishing prompt should never be shown to the user. (So, really, the user should not have any visibility into this.)

aips/aip-x.md
```diff
@@ -0,0 +1,707 @@
...
+- Replay attacks — A hacker must not be able to use a signature outside of its intended usage.
+- Forgery attacks — A hacker must not be able to provide a false signature that can bypass verification.
+
+## Future Potential
+
+This AIP formally introduces a standard for authentication via Aptos decentralized identities. In the future, the standard may be extended to provide additional information from the wallet.
```
alinush, line 688: > "via Aptos decentralized identities."

This is not precise. Unclear what "decentralized identity" means. It's a fluff umbrella term, AFAICT.

"via Aptos accounts" would be precise.

aips/aip-x.md
```diff
@@ -0,0 +1,707 @@
...
+
+With the rise of Keyless on the Aptos Blockchain, wallet connections may want to be extended to provide additional personal information about the user, this would include full names, phone numbers, or emails.
```
alinush, line 690: This could actually be made possible using the `extra_field` feature of the keyless ZKP (see [AIP-61](https://github.com/aptos-foundation/AIPs/blob/main/aips/aip-61.md)).

aips/aip-x.md
```diff
@@ -0,0 +1,707 @@
...
+
+The "Sign in with Aptos" (SIWA) standard introduces a secure and user-friendly way for users to authenticate to off-chain resources by proving ownership of their blockchain addresses. It simplifies the authentication process by replacing the traditional `connect` + `signMessage` flow in the wallet standard with a streamlined one-click `signIn` method. SIWA leverages decentralized identities to avoid reliance on traditional schemes like SSO while incorporating security measures to combat phishing attacks and improve user visibility.
+
+### Out of scope
+
+The standard relies on public key signature verification. This AIP will not cover specifications or correctness of signature verification.
+
+## High-level Overview
+
+**Decentralized authentication** is a privacy-first solution that empowers individuals to safeguard their information, personalize their experiences, and maintain freedom from censorship.
```
alinush, line 23: I don't think that's what this AIP is about... there are no privacy benefits here over the `connect` + `signIn` baseline, right? 

This AIP is about users signing into a dapp using their Aptos Account and getting access to protected resources.

"Decentralized authentication" sounds like a vague umbrella term to me that should be avoided.

aips/aip-x.md
```diff
@@ -0,0 +1,707 @@
...
+| fields.notBefore      | ISO 8601 string |           | `date-time` that indicates when the signed authentication message starts being valid.                                                 |
+| fields.requestId      | string          |           | System-specific identifier used to uniquely refer to the authentication request.                                                      |
+| fields.resources      | string[]        |           | List of information or references to information the user wishes to have resolved as part of the authentication by the relying party. |
+| plainText             | string          | ✓         | The constructed message using the standardized message format.                                                                        |
+| signature             | bytes           | ✓         | The signature of the message. This must conform to the domain separated signing algorithm.                                            |
+| type                  | string          | ✓         | The public key scheme used to sign the message (e.g. 'ed25519', 'single_key', 'multi_key', 'multi_ed25519', etc.)                     |
+
+### AptosSignInOutput
```
alinush, line 120: I don't like this generic "output" name. Shouldn't this be something more like a `AptosSignInSignature` or `SiwaSignature`?

aips/aip-x.md
alinush: I should've also highlighted the things you did very well when writing this AIP:
- Well-structured
- Writing is generally good
- Figures / screenshots / illustrations 👌👌👌
- Nice tables describing the formats
- You included a formal description of the grammar
- Code examples
- Lots of examples in general

aips/aip-x.md
alinush: Can you point via URL to AIP-62's `*MessageInput` structs with the optional `application` field, for extra clarity.

aips/aip-x.md
```diff
@@ -0,0 +1,707 @@
...
+
+### Domain Separating Signing Algorithm
+
+A plaintext construction of a SIWA message must be prefixed with a valid SIWA hash before signing to ensure domain separation across other signatures. The messages must conform to the following expression:
+
+```
+sha3_256(b"SIGN_IN_WITH_APTOS::") || <siwa_message>
+```
+
+### Wallet Standard
```
alinush, line 236: Retitle: Modifications to the wallet standard

aips/aip-x.md
```diff
@@ -0,0 +1,707 @@
...
+The features below will be added to the wallet as suggested features. Features will be optional to maintain compatibility with existing wallets.
+
+`aptos:signIn` method to establish a connection and to sign a _Sign in with Aptos_ authentication message.
+
+```ts
+// `input: AptosSignInInput` - inputs with fields to construct an authentication message
+signIn(input: AptosSignInInput): Promise<UserResponse<AptosSignInOutput>>
+```
+
+### Wallet Implementer
```
alinush, line 251: Retitle: How to implement this AIP safely as a wallet developer

aips/aip-x.md
alinush: After convo: You mean "signature verification" (which takes a message `m`, a public key `pk` and a signature `\sigma` as input and outputs either 0 or 1)

aips/aip-x.md
alinush: Regarding #1:

I think the only opportunities to mess this up are:
- the application does not pick the `nonce` randomly (but even this the TS SDK could fix, by not even letting the application pick the nonce, no?)
- the session management via cookies

PR #558, opened 2025-02-01
AIP-X Increase coin symbol limit to 32 bytes

gregnazario:


aips/aip-increase-coin-symbol.md
```diff
@@ -0,0 +1,68 @@
...
+created: 02/01/2024
+updated (*optional): <mm/dd/yyyy>
+requires (*optional): <AIP number(s)>
+---
+
+# AIP-X - Increase Coin Symbol Byte Length
+
+## Summary
+
+Currently, emojis are 8-bytes, and could be longer.  This AIP proposes to increase the coin symbol byte length to allow
```
alnoki, line 18: > emojis are 8-bytes

This is factually incorrect, counterexample (3 bytes): ⏰ https://github.com/econia-labs/emojicoin-dot-fun/blob/902b89a3b8c9606208eda65118bbe9c4d0a58ed5/src/move/emojicoin_dot_fun/sources/hex_codes.move#L23

gregnazario: Okay, this is a technicality typo... Where the "max" is 8 bytes correct?

alnoki: There are some that are more than 8 bytes, for example the Scottish flag 🏴󠁧󠁢󠁳󠁣󠁴󠁿 is 28 bytes...:

https://github.com/econia-labs/emojicoin-dot-fun/blob/f2db7f1ffb2107b792f3d94c30b1c2d96d26b66b/src/move/emojicoin_dot_fun/sources/hex_codes.move#L2405

gregnazario: What do you propose is the longest then?

alnoki: > the longest

There are emojis that are even longer than 28 bytes, and we had to take some of them out to keep package under size limit: https://github.com/econia-labs/emojicoin-dot-fun/blob/f2db7f1ffb2107b792f3d94c30b1c2d96d26b66b/src/python/move_emojis/utils/data_parser.py#L203-L220

Also pretty much any combination that combines modifiers like skin color or man/woman (especially for the 4-person families) can quickly add up

FELLIX000 (commented):
Please help me understand this

aips/aip-increase-coin-symbol.md
gregnazario: Yeah, I suspect 32-bytes then should be coverage enough of most emojis, plus a decent number of combinations of simple emojis

MoonShiesty:
max FA symbol length will need to be increased as well

FELLIX000 (approved):

gregnazario:
> max FA symbol length will need to be increased as well

This is increased here too

sherry-x (approved):

thepomeranian (approved):

PR #561, opened 2025-02-17
aip: add signer capability support for object code deployment

mshakeg:
## Description

Adds AIP proposal for enabling signer capability support in object code deployment. This AIP outlines a solution to enable modules published under objects to access their object signer during initialization and store it for later use, similar to how modules published under resource accounts can retrieve and store their SignerCapability.

### DIFF
--- a/aips/object-code-deployment-auth-ref.md
+++ b/aips/object-code-deployment-auth-ref.md
@@ -65,7 +65,7 @@ The implementation adds three key components to the `object_code_deployment` mod
 1. AuthRef struct:
 ```move
 /// Authorization reference for an object that has code published to it
-struct AuthRef has store {
+struct AuthRef has drop, store {
     object_address: address
 }
 ```
--- a/aips/object-code-deployment-auth-ref.md
+++ b/aips/object-code-deployment-auth-ref.md
@@ -1,0 +1,167 @@
+---
+aip:
+title: Add signer capability support for object code deployment
+author: mshakeg
+discussions-to:
+Status: Draft
+type: Standard (Framework)
+created: 02/17/2024
+---
+
+# AIP-X - Add signer capability support for object code deployment
+
+## Summary
+
+When publishing a package under an object, modules cannot access their object signer after initialization, making it impossible to manage funds sent to the object's primary fungible store. This AIP proposes adding AuthRef support to enable modules to securely access their object signer during initialization and store it for later use, similar to how modules published under resource accounts can retrieve and store their SignerCapability.
+
+The solution introduces an AuthRef abstraction that can only be generated during module initialization, maintaining security while enabling proper fund management capabilities for object-published modules.
+
+### Out of scope
+
+- Changes to how packages are published under objects
+- Changes to existing package upgrade mechanisms
+- Additional permissioned functions for object management
+- Changes to modules already published under objects (cannot retroactively gain AuthRef access)
+
+## High-level Overview
+
+The proposal adds AuthRef support to the `object_code_deployment` module, enabling modules published under objects to generate an AuthRef during initialization that can later be used to generate the object's signer. This brings object-based publishing closer to parity with resource account-based publishing, where modules can already retrieve and store their SignerCapability during initialization.
+
+The AuthRef serves as a capability, proving that a module has the right to generate its object's signer. It can only be created during module initialization when the object signer is naturally available, following Move's capability-based security model.
+
+## Impact
+
+This change impacts:
+- Developers publishing modules under objects who need programmatic control over their module's resources
+- Existing modules published under objects (cannot retroactively gain AuthRef access)
+- Future module designs that may choose object-based publishing over resource account-based publishing
+
+Without this change, modules published under objects would remain limited in their ability to manage funds and resources autonomously, potentially forcing developers to use resource accounts instead or implement complex workarounds.
+
+## Alternative Solutions
+
+1. Restricted API Approach:
+```move
+pub fun transfer<CoinType>(owner: &signer, amount: u64, destination: address);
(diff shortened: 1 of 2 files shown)


lightmark (commented):
struct AuthRef has store {
    object_address: address
}

maybe it's called `extend_ref_accessor`. btw, why doesn't it have `drop`

mshakeg:
@lightmark are you saying I should rename the `object_address` field in the `AuthRef` struct to `extend_ref_accessor`?

I can add the `drop` ability to `AuthRef`, it would be consistent with `account::SignerCapability`

gregnazario (approved):

aips/object-code-deployment-auth-ref.md
```diff
@@ -0,0 +1,167 @@
...
+## Specification and Implementation Details
+
+The implementation adds three key components to the `object_code_deployment` module:
+
+1. AuthRef struct:
+```move
+/// Authorization reference for an object that has code published to it
+struct AuthRef has store {
+    object_address: address
+}
```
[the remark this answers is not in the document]
mshakeg: @gregnazario ok, I added `drop`

aips/object-code-deployment-auth-ref.md
```diff
@@ -0,0 +1,167 @@
...
+- Changes to how packages are published under objects
+- Changes to existing package upgrade mechanisms
+- Additional permissioned functions for object management
+- Changes to modules already published under objects (cannot retroactively gain AuthRef access)
+
+## High-level Overview
+
+The proposal adds AuthRef support to the `object_code_deployment` module, enabling modules published under objects to generate an AuthRef during initialization that can later be used to generate the object's signer. This brings object-based publishing closer to parity with resource account-based publishing, where modules can already retrieve and store their SignerCapability during initialization.
+
+The AuthRef serves as a capability, proving that a module has the right to generate its object's signer. It can only be created during module initialization when the object signer is naturally available, following Move's capability-based security model.
```
movekevin, line 30: Thanks for starting this AIP! Can I ask what’s the rationale for adding this an auth ref that would require explicit init module template code vs just adding a function to allow arbitrary access to signer as long as that function is called from the same address (or other similar options with more safeguards)? The approach in AuthRef seems longer and less convenient to use for developers. They'd have copy paste code for init module (similar to https://github.com/aptos-labs/aptos-core/tree/main/aptos-move/move-examples/package_manager) to ensure they don't lose access later. Having a more direct way would be better dev ex as people don’t need to remember and existing code can get auto access as long as they’re upgradable

aips/object-code-deployment-auth-ref.md
```diff
@@ -0,0 +1,167 @@
...
+
+## Risks and Drawbacks
+
+1. Backwards Compatibility:
+- Existing modules published under objects cannot retroactively gain AuthRef access
+- Mitigation: Document limitation clearly, only affects future deployments
+
+2. Security Considerations:
+- Modules have full signer access after storing AuthRef
+- Mitigation: AuthRef can only be generated during initialization, following established capability patterns
```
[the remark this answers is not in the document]
mshakeg: @cfal you are correct, though only the object signer is allowed to call `generate_auth_ref` so I don't really see the issue?

https://github.com/aptos-labs/aptos-core/pull/15954/files

JohnChangUK:
Great idea, some questions:

1. What about having the owner of the code object or RBAC roles to have permissions to manage `AuthRef`? 
2.  How do you plan on tracking who has access to the `AuthRef`? 
    - Would need some events at the very least.
3. What if I want to revoke access to people who have `AuthRef`?
4. What if I want to renounce ownership of the contract?

mshakeg:
@JohnChangUK I'm thinking of making a change to my current `generate_auth_ref` implementation to limit it to only 1 AuthRef for a given object signer, this way it's implicitly forced to only be called in one `init_module` function for the pkg published under an object.

The AuthRef is meant for the pkg to have access to its signer programmatically so if the pkg owner renounces ownership that would have no impact. Also why would the pkg owner need to manage the `AuthRef`?

aips/object-code-deployment-auth-ref.md
JohnChangUK: An issue with deploying to resource accounts is that if you don't deploy `package_manager` along with your contracts, you forfeit access to the `signer`. With the `AuthRef` having to be created during construction, it faces a similar problem where if you do not have this logic in your contract , you lose access to the `signer`.

One option could be to have the owner of the code object generate `AuthRef`, something similar to the collection owner having access to the collection object `signer` [here](https://github.com/aptos-labs/token-minter/blob/main/token-minter/sources/modules/collection_components.move#L249).
This means if they forget, they can call a function to generate the `AuthRef` and store it at a later stage.

mshakeg: @JohnChangUK when a pkg is upgraded is `init_module` called again? Or is it only called once at the time of initial deployment? If it's the latter then I agree it could be useful to extend the `generate_auth_ref` to also accept the signer of the owner of the pkg object, however if it's the former then the pkg could be upgraded to have `init_module` called again to have the module call `generate_auth_ref` in this upgraded `init_module`?

JohnChangUK: Upgrades do not invoke the init module function from what I remember.

mshakeg: @JohnChangUK ok, then I agree that `generate_auth_ref` should also allow for the pkg object owner to generate the `AuthRef`

movekevin: I'm thinking there doesn't need to be an "owner" defined. Basically there will be this function

```
public fun package_signer(): signer {}
```

that's only callable from the same address

thepomeranian (approved):

PR #572, opened 2025-03-18
[AIP-116] Update aip-116.md with revisions

aptqpv:
### Description
1. Removed "wallet journey" examples from the motivation summary
2. Re-ordered examples in motivation section
3. Fixed incorrect section references
4. Removed `signingMessage` and `input` from `AptosSignInOutput` and replaced `plainText` with `message`
5. Added a section on "Resource Injection"
6. Added a section on message and signature verification
7. Added implementation examples for message and signature verification in TypeScript
8. Added rationale in for display error messages in the "Displaying Errors vs Rejecting Requests" section


aips/aip-116.md
```diff
@@ -26,29 +26,25 @@ Authenticating with Aptos accounts is a privacy-first solution that empowers ind
...
-- **Petra** – _Passwords/Biometric_: - Fully non-custodial. Requires no information to be shared with Petra, ensuring complete privacy and resistance to censorship.
-- **Aptos Connect** – _Social Providers_: Fully non-custodial. Shares limited information with Aptos Connect including name, email, `sub` and may be subject to censorship by social providers.
+The `connect` + `signMessage` flow...
 
-However, while this approach offers significant benefits, the existing `connect` + `signMessage` flow used for authentication presents a few challenges:
+![Figure 1: Connect and Sign Message Flow](./static/10.png)
+
+- **presents security risks —** Malicious websites can deceive users into signing messages intended for legitimate dApps.
+- **is unintuitive —** The traditional `connect` + `signMessage` workflow involves two separate steps, creating a clunky and unintuitive user experience.
+- **trains the user to sign malicious messages —** Users are being trained to sign messages without understanding its contents. The long-term implications of this practice can lead to users becoming more susceptible to phishing attacks.
```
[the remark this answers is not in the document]
alinush: "The long-term implications of this practice" --> "Long term, this"

aips/aip-116.md
```diff
@@ -26,29 +26,25 @@ Authenticating with Aptos accounts is a privacy-first solution that empowers ind
 
 By authenticating using Aptos accounts, users have full control over their authentication journey. They can decide _what to share_ and _how to log in,_ ensuring a more secure and customizable experience.
```
alinush, line 27: This paragraph and the one above it are fluff (they are not saying anything meaningful). I think you should remove them and just start with what you wrote below "The existing `connect` + `signMessage` flow"

aips/aip-116.md
```diff
@@ -26,29 +26,25 @@ Authenticating with Aptos accounts is a privacy-first solution that empowers ind
...
 <br/>
 
-And the `connect` + `signMessage` flow...
-
-![Figure 2: Connect and Sign Message Flow](./static/10.png)
-
-- **presents security risks —** Malicious websites can deceive users into signing messages intended for legitimate dApps.
-- **is unintuitive —** The traditional `connect` + `signMessage` workflow involves two separate steps, creating a clunky and unintuitive user experience.
-
 Addressing these challenges is key to unlocking the full potential of Aptos accounts and delivering a seamless, secure, and user-friendly experience.
```
alinush, line 48: Unnecessary fluff: I would remove.

You've already made a strong point above.

aips/aip-116.md
```diff
@@ -91,7 +87,7 @@ The new standard introduces new interfaces that conform to [CAIP-122](https://ch
 ### Definitions
 
 - _binding_: The process of adding or modifying fields to fit requirements, sourcing fields from a trusted data source such as the browser window. This process will typically be done by the wallet.
-- _SIWA Message_: A plain text message that is constructed from the `AptosSignInInput` fields by the wallet. This message should follow the format specified in the [Message Construction](#message-construction) section.
```
alinush, line 94: I would not capitalize "Message" here, since you're gonna use it in lowercase as "message" throughout the text. Same for others.

aips/aip-116.md
```diff
@@ -120,22 +116,22 @@ The `domain` field may be omitted from the `AptosSignInInput` to prevent the cre
...
-| signatureType     | string             | ✓         | The key scheme used to sign the message (e.g. 'ed25519', 'single_key', 'multi_key', 'multi_ed25519', etc.).                                                                                                                                           |
+| Name              | Type   | Description                                                                                                        |
+| ----------------- | ------ | ------------------------------------------------------------------------------------------------------------------ |
+| account.address   | string | The address of the user that is signing in. This field MUST match the `address` field used to construct `message`. |
+| account.publicKey | bytes  | The public key of the user that is signing in.                                                                     |
+| message           | string | The constructed message using the standardized message format.                                                     |
```
alinush, line 123: Be specific & use your own terminology: Is this a SIWA message or SIWA signing message? If the latter, maybe rename this field to `signingMessage` for clarity.

aips/aip-116.md
```diff
@@ -120,22 +116,22 @@ The `domain` field may be omitted from the `AptosSignInInput` to prevent the cre
...
+| signature         | bytes  | The signature of the signing message.                                                                              |
+| signatureType     | string | The key scheme used to sign the message (e.g. 'ed25519', 'single_key', 'multi_key', 'multi_ed25519', etc.).        |
```
alinush, line 125: --> "The digital signature scheme used to sign (e.g., [...])"

aips/aip-116.md
alinush, line 1: `**AptosSignInInput**` and `**AptosSignInOutput**` --> **`AptosSignInInput`** and **`AptosSignInOutput`**

alinush: At some point you say this "ensures compatibility across applications" --> But how? It is not immediately clear to me what this means & you would have to clarify (or remove if fluff).

alinush: "By enabling wallets to validate and flag potentially malicious messages, SIWA minimizes the risk of phishing attacks and other security vulnerabilities." --> Actually, it eliminates it, not just reduces it right?

aips/aip-116.md
```diff
@@ -120,22 +112,22 @@ The `domain` field may be omitted from the `AptosSignInInput` to prevent the cre
```
alinush, line 112: A little bit confused by the comment above:

> "The `domain` field may be omitted from the `AptosSignInInput` to prevent the creation of application-specific phishing messages."

This field should **never** be omitted, no? And that's how you prevent phishing. (Not sure it makes sense to distinguish "phishing" as "application-specific"...)

aips/aip-116.md
```diff
@@ -91,8 +83,8 @@ The new standard introduces new interfaces that conform to [CAIP-122](https://ch
 ### Definitions
```
alinush, line 83: Above, in "## Specification and Implementation Details", you should **sketch what happens in SIWA at a high level** and how these structs are used:
- the user goes to a dapp and sees "sign in with aptos"
- clicks the button
- the dapp constructs a signing request (?) for the wallet, in the form of a partially-filled-in `AptosSignInInput` detailing what dapp this is (is that right?)
- the wallet fills in the rest of details, updating the `AptosSignInInput` object
- the wallet displays the SIWA prompt to the user, asking them to sign into the dapp (effectively summarizing what's in the `AptosSignInInput` object)
- the wallet serializes this object and signs it, producing an `AptosSignInOutput` object
- the dapp obtains this signature from the wallet and verifies it
- if it verifies, the dapp is certain that it authenticated the user with the Aptos address specified in the `AptosSignInOutput` object and can start a session for them and give them access to their resources

I think it's important to summarize this (or, if incorrect, summarize the correct thing), since it helps everyone (including me) understand what messages are flying around from who to whom, and how this whole thing works 👍

aips/aip-116.md
```diff
@@ -91,8 +83,8 @@ The new standard introduces new interfaces that conform to [CAIP-122](https://ch
...
 
 - _binding_: The process of adding or modifying fields to fit requirements, sourcing fields from a trusted data source such as the browser window. This process will typically be done by the wallet.
-- _SIWA Message_: A plain text message that is constructed from the `AptosSignInInput` fields by the wallet. This message should follow the format specified in the [Message Construction](#message-construction) section.
-- _SIWA Signing Message_: Bytes output from the [Domain Separated Signing Algorithm](#domain-separated-signing-algorithm) using the SIWA message as input.
+- _SIWA message_: A plain text message that is constructed from the `AptosSignInInput` alongside bound fields by the wallet. This message should follow the format specified in the [Message Construction](#siwa-message-construction) section.
+- _SIWA signing message_: Bytes output from the [Domain Separated Signing Algorithm](#domain-separated-signing-algorithm) using the SIWA message as input.
 - _Signing Request_: A request from an application to the wallet to conduct a certain operation such as signing a message.
 
 ### AptosSignInInput
```
alinush, line 90: What I would like this to explain is what does it mean for a field to be "mandatory".

AFAICT, it means that the dapp has to fill it in before sending the signing request to the wallet

aips/aip-116.md
```diff
@@ -120,22 +112,22 @@ The `domain` field may be omitted from the `AptosSignInInput` to prevent the cre
...
+| account.publicKey | bytes  | The public key of the user that is signing in.                                                                     |
+| message           | string | The constructed SIWA message using the standardized message format.                                                |
+| signature         | bytes  | The signature of the signing message.                                                                              |
+| signatureType     | string | The digital signature scheme used to sign (e.g. 'ed25519', 'single_key', 'multi_key', 'multi_ed25519', etc.).      |
 
 If there are library representations of the `account.publicKey` and `signature`, they may be used instead of the `bytes` type (e.g. `PublicKey` and `Signature` from the `@aptos-labs/ts-sdk` package).
 
 ### SIWA Message Construction
 
 The proposed plain text representation will be an adaptation of the standards specified from [EIP-4361](https://eips.ethereum.org/EIPS/eip-4361) and [CAIP-122](https://chainagnostic.org/CAIPs/caip-122). If certains fields are not present in the `AptosSignInInput`, omit them from the SIWA message.
```
alinush, line 127: The proposed plain text representation **of an `AptosSignInInput` object**

aips/aip-116.md
```diff
@@ -120,22 +112,22 @@ The `domain` field may be omitted from the `AptosSignInInput` to prevent the cre
...
 
+If the following fields are not present in the `AptosSignInInput`, they must be bound by the wallet: `address`, `chainId`, `domain`, `uri`, and `version`.
```
alinush, line 129: If the user has multiple addresses in his account, should this AIP also cover how the user may be able to switch his address to another one after seeing the sign-in prompt and noticing its the wrong address?

Or is that left as an implementation detail for wallet devs reading this?

aips/aip-116.md
alinush: "This is example occurs" --> "This example occurs"

alinush: After "A plain text construction of a SIWA message", add "(as described [here](#SIWA-Message-Construction))"

aips/aip-116.md
```diff
@@ -242,6 +234,40 @@ sha3_256(b"SIGN_IN_WITH_APTOS::") || <siwa_message>
...
+
+- `publicKey` - The public key of the user that is signing in. Should be the `publicKey` used in the `signature` verification step above.
+- `expected` - The `AptosSignInInput` fields that are expected to be present in the `message`.
+- `message` - The SIWA message to verify.
+
+The verification process must perform the following steps:
+
+1. Parse the `message` into an `AptosSignInInput` object.
+2. Compare the `expected` fields against the parsed `AptosSignInInput` object. If fields are present in the `expected` object, they must match the corresponding fields in the parsed `AptosSignInInput` object.
+3. Ensure that the `publicKey`'s authentication key's originating address is equal to the `address` field in the parsed `AptosSignInInput` object.
```
alinush, line 266: Why does this AIP care above the originating address table, which is deprecated anyway?

Instead of 3. and 4. here, did you mean to simply say:
First, derive the authentication key `authKey` of `publicKey` and check if it is equal to `address`. If not equal, check that, on-chain, the `authentication_key` field of the Aptos `Account` resource for `address` is set to `authKey`.

(Although, this does not account for future changes to `Account[V2?]`, but it can't anyway.)

aips/aip-116.md
```diff
@@ -242,6 +234,40 @@ sha3_256(b"SIGN_IN_WITH_APTOS::") || <siwa_message>
...
+4. If there is no originating address, ensure that the `publicKey`'s authentication key is equal to the `address` field in the parsed `AptosSignInInput` object.
+
+Implementation examples for both signature and message verification can be found in the [Verification Reference Implementations](#verification-reference-implementations) section.
+
 ### Modifications to the Wallet Standard
 
 This AIP also extends the [Wallet Standard](https://github.com/aptos-labs/wallet-standard), which is a chain-agnostic set of interfaces and conventions that aim to improve how applications interact with injected wallets.
```
alinush, line 273: Isn't this the **Aptos** wallet standard that is being extended? Which is **not** actually chain agnostic...

aips/aip-116.md
alinush: I still don't understand what you are trying to say here:
> The interface for the signing request sent to the wallet may have a variant of the `AptosSignInInput` interface that omits the `domain` field to prevent the creation of application-specific phishing messages.

aips/aip-116.md
alinush: Bump.

aips/aip-116.md
alinush: Bump. I don't think you addressed this.

alinush: Oh, you did do it but you did it in more depth below rather than here. That's nice and still useful, but I was trying to suggest that without a quick sketch of how these structs are used here, the more in depth discussion below is hard to follow.

alinush: (e.g., like the `mandatory` fields)

aips/aip-116.md
```diff
@@ -114,27 +106,42 @@ The `AptosSignInInput` interface will be included in the **signing request** to
...
+2. The application will make a call to the backend to construct an `AptosSignInInput` object. This object can be partially filled, only requiring the fields that have been marked mandatory in the [`AptosSignInInput`](#aptossignininput) table above. This partially filled `AptosSignInInput` object should be stored to be verified later in step 6.
+3. Once the application receives the `AptosSignInInput` object, the application will make a signing request to the wallet containing the `AptosSignInInput` object. This will prompt the user to a sign a SIWA message.
+4. The wallet will construct a SIWA message from the `AptosSignInInput`, binding required fields that are not present in the partially filled `AptosSignInInput` object. Once the SIWA message is constructed, the SIWA message will go into the [Domain Separated Signing Algorithm](#domain-separated-signing-algorithm) to create a SIWA signing message. The prompt will display the SIWA message in a UI that can easily be read by the user.
+5. When the user approves to sign the message, the wallet will use the Aptos account will be used to sign the SIWA signing message and return an `AptosSignInOutput` object to the application.
```
alinush, line 133: Typo:
> the wallet will use the Aptos account will be used to sig

alinush: "The wallet will use the Aptos account to sign"?

aips/aip-116.md
```diff
@@ -234,17 +241,50 @@ message-resources = \\*( LF "- " URI )
...
-A plain text construction of a SIWA message must be prefixed with a `SIGN_IN_WITH_APTOS::` hash before signing to ensure domain separation across other signatures. The SIWA signing messages must conform to the following expression:
+A plain text construction of a SIWA message (as described [here](#siwa-message-construction)) must be prefixed with a `SIGN_IN_WITH_APTOS::` hash before signing to ensure domain separation across other signatures. The SIWA signing messages must conform to the following expression:
 
 ```
 sha3_256(b"SIGN_IN_WITH_APTOS::") || <siwa_message>
 ```
 
 where `<siwa_message>` is the plain text construction of a SIWA message explained in the [Message Construction](#message-construction) section.
 
+### Signature and Message Verification
```
alinush, line 252: Can we have a proper term for this backend SIWA verification? Maybe we can call it "**Backend SIWA verification**"?

aips/aip-116.md
```diff
@@ -234,17 +241,50 @@ message-resources = \\*( LF "- " URI )
...
+The verification process consists of two equally important components:
+
+1. **Signature verification** - Cryptographically ensures the authenticity of a signature against the `message`.
+2. **SIWA Message verification** - Ensures the content of the `AptosSignInOutput` message adheres to the expected fields.
+
+#### Signature Verification
+
+Signature verification must consist of three parts:
+
+- `publicKey` - The public key of the user that is signing in.
```
alinush, line 263: May be helpful to further clarify: of the **Aptos** user that is signing in.

aips/aip-116.md
```diff
@@ -234,17 +241,50 @@ message-resources = \\*( LF "- " URI )
...
+- `message` - The SIWA message to verify.
+
+To verify the `signature`, the `publicKey` should be used to cryptographically verify the `signature` against the `message`. This verification should be stateless.
+
+#### Message Verification
+
+The message verification process must consist of the following parts:
+
+- `publicKey` - The public key of the user that is signing in. Should be the `publicKey` used in the `signature` verification step above.
+- `expected` - The `AptosSignInInput` fields that are expected to be present in the `message`.
```
alinush, line 274: There is a problem here for me: I have no idea what should be in this `expected` fields.

aips/aip-116.md
```diff
@@ -265,7 +305,38 @@ To ensure the safety of our users, we want to take a conservative approach when
...
+- Wallet implementors MUST ensure that the bound fields (`address`, `chainId`, `domain`, `uri`, and `version`) match their corresponding fields in the `AptosSignInInput` _if present_. If the bound fields do not match the fields presented in the signing request, errors must be displayed to the users detailing the issues.
+- In the case that the signing Aptos account was rotated, the bound `address` field MUST be the originating address of the authentication key.
+
+### Wallet Interactions
+
+- Prior to signing, the wallet MAY consult users for preferences, such as the selection of one `address` out of many or the `network` to use.
+
+#### Resource Injection
+
+Wallets may inject additional information into the `resources` field for **wallet specified resources** and in any **wallet specified formats**.
```
alinush, line 317: I think the reader should be reminded when & where this injection happens: is it when the wallet gets the `AptosSignInInput` message and fills it in?

aips/aip-116.md
```diff
@@ -326,47 +395,63 @@ A new npm package named `@aptos-labs/siwa` will be created to house utilities fo
...
  *
- * @param input The input fields to create the SIWA plain text message from.
+ * @param input The `AptosSignInInput` with required fields to create the SIWA message from.
  *
- * @returns The SIWA plain text message.
+ * @returns The SIWA message.
  */
-function createSignInMessagePlainText(
+function createSignInMessageText(
   input: AptosSignInInput & AptosSignInRequiredFields
```
alinush, line 411: I think this `AptosSignInRequiredFields` appears out of nowhere and is not explained...

aips/aip-116.md
alinush: Oh, I think I understand but only cause I found this later on below:

> The domain field has been omitted from the AptosSignInInput interface to prevent the creation of application-specific phishing messages.

I am still rather confused though: if you don't have a domain field there anymore, then your verification logic must take the domain separately... which is a bit ugly. I will look below to understand.

aips/aip-116.md
```diff
@@ -326,47 +395,63 @@ A new npm package named `@aptos-labs/siwa` will be created to house utilities fo
...
+ *
+ * @param output The `AptosSignInOutput` to verify against the input.
+ *
+ * @returns The `AptosSignInInput` fields that are parsed from the message.
+ */
+function verifySignInSignature(output: {
+  publicKey: PublicKey;
+  signature: Signature;
+  message: string;
+}): VerificationResult<AptosSignInInput & AptosSignInRequiredFields>;
```
alinush, line 436: This is rather odd to me: typically signature verification returns `true` / `false`.

Furthermore, deserializing during signature verification is an anti-pattern. It's more secure to verify a signature over a structured object by (1) serializing it and (2) checking the signature on the serialization (+ whatever domain separation is needed).

aips/aip-116.md
```diff
@@ -114,27 +111,49 @@ The `AptosSignInInput` interface will be included in the **signing request** to
...
-| signature         | bytes              | ✓         | The signature of the signing message.                                                                                                                                                                                                                 |
-| signatureType     | string             | ✓         | The key scheme used to sign the message (e.g. 'ed25519', 'single_key', 'multi_key', 'multi_ed25519', etc.).                                                                                                                                           |
+| Name              | Type             | Description                                                                                                        |
+| ----------------- | ---------------- | ------------------------------------------------------------------------------------------------------------------ |
+| account.address   | string           | The address of the user that is signing in. This field MUST match the `address` field used to construct `message`. |
```
alinush, line 129: It is very unclear what `message` is referring to here, especially since it is quoted with `` .

aips/aip-116.md
```diff
@@ -114,27 +111,49 @@ The `AptosSignInInput` interface will be included in the **signing request** to
...
+| account.publicKey | bytes            | The public key of the Aptos user that is signing in.                                                               |
+| input             | AptosSignInInput | The `AptosSignInInput` with bound fields added by the wallet.                                                      |
+| signature         | bytes            | The signature of the signing message.                                                                              |
```
alinush, line 132: *Signature **over** the SIWA signing message.

aips/aip-116.md
```diff
@@ -234,28 +253,53 @@ message-resources = \\*( LF "- " URI )
...
+- `input` - The `AptosSignInInput` object with bound fields that was used to construct a SIWA signing message and signed to produce the `signature`.
+
+To verify the `signature`, the `publicKey` should be used to cryptographically verify the `signature` against the `message` and outputs `true` or `false` to indicate success or failure. This verification should be stateless.
+
+#### Message Verification
+
+Message verification takes three parts:
+
+- `publicKey` - The public key of the user that is signing in. Should be the `publicKey` used in the `signature` verification step above.
+- `expected` - The `AptosSignInInput` object that are expected to be present and match the `input` object.
```
alinush, line 286: Fix grammar: "object that are expected"

aips/aip-116.md
```diff
@@ -234,28 +253,53 @@ message-resources = \\*( LF "- " URI )
...
+- `input` - The `AptosSignInInput` object with bound fields that was used to construct the `input`.
```
alinush, line 287: `input` seems to be defined recursively in terms of itself here.

alinush (approved):
✅

PR #578, opened 2025-04-01
Create aip-order-book

igor-aptos:


fikunmi-ap:
why are you implementing parts in Move and not the whole hog?

igor-aptos:
@fikunmi-ap sorry, missed your comment. Generally move code is more cleaner to write and is more maintainable. 
So we will implement things natively where it makes sense. The biggest usefulness we see of natives here would be to work better with BlockSTM and provide better parallelism characteristics.

Any reason you would like to see everything be native?

fikunmi-ap:
Performance.

The less compute you have to do in the VM, the better. It directly translates to orders/second.

I primarily work on the SVM and I believe it's currently no where near performant enough for an order book. The MoveVM is currently even slower.

So that's it 😄

zekun000 (approved):

aips/aip-order-book.md
```diff
@@ -0,0 +1,326 @@
+---
+aip: (this is determined by the AIP Manager, leave it empty when drafting)
```
davidiw, line 2: just take the next number available so that it is globally referenceable and we don't end up with a dead link

aips/aip-order-book.md
```diff
@@ -0,0 +1,326 @@
...
+      of the order when its placed. The clearinghouse is expected to validate the order and return the maximum settlement size
+      of the order. Checkout clearinghouse_test as an example of the simplest form of clearing house implementation that just tracks
+      the position size of the user and does not do any validation.
+## Impact
+
+- Anyone who wants to build a high-throughput DEX
+
+## Alternative Solutions
+
+Keeping GTE on the ecosystem to implement, might not allow for all the optimizations that can be done with tight integration with the VM/BlockSTM.
```
davidiw, line 76: I think the right wording here is more on pure Move versus the hybrid solution.

aips/aip-order-book.md
```diff
@@ -0,0 +1,326 @@
...
+
+Following are the main APIs supported by the CLOB
+
+```
+   struct OrderBook<M: store + copy + drop> has store;
+
+    struct SingleOrderMatch<M: store + copy + drop> has drop, copy {
+        order: Order<M>,
+        matched_size: u64
+    }
```
davidiw, line 88: curious why we didn't use objects

davidiw (commented):
missing events api?

aips/aip-order-book.md
sitalkedia: Can you elaborate.. how do you envision we using objects here?

PR #580, opened 2025-04-08
Update DAA text

igor-aptos:


aips/aip-113.md
```diff
@@ -11,30 +11,35 @@ created: 02/14/2024
...
 How to properly design an derivable AA authentication function/module is out of scope.
 
 ## High-level Overview
-In Derivable AA, authentication is scoped to a domain. You can register an authentication function, which then creates a domain. The authentication logic is not tied to a specific account but instead relies on the domain and account identity. Here's how it works:
+In Derivable AA, registering an authentication function defines a collections of accounts being authenticated by it. Here's how it works:
 
-1. The user provides an `account_identity`, which is included in the authentication data (`auth_data`).
-2. Based on the `account_identity` and the associated `function_info`, the system derives an account address for authentication.
+1. The user provides an `abstrat_public_key`, which is included in the authentication data (`auth_data`).
+2. Based on the `abstrat_public_key` and the associated `function_info`, the system derives an account address for authentication.
```
hardsetting, line 27:
```suggestion
1. The user provides an `abstract_public_key`, which is included in the authentication data (`auth_data`).
2. Based on the `abstract_public_key` and the associated `function_info`, the system derives an account address for authentication.
```

aips/aip-113.md
```diff
@@ -45,67 +50,66 @@ Derivable AA introduces the following features:
...
-): address {
-    let function_info = function_info::new_function_info_from_address(module_address, module_name, function_name);
-    let bytes = bcs::to_bytes(&function_info);
-    bytes.append(bcs::to_bytes(account_identity));
-    bytes.push_back(DOMAIN_ABSTRACTION_DERIVED_SCHEME);
-    from_bcs::to_address(hash::sha3_256(bytes))
-}
+    /// Return the account address corresponding to the given `abstract_public_key`,
+    /// for the derivable account abstraction defined by the given function.
+    public fun derive_account_address(derivable_func_info: FunctionInfo, abstract_public_key: &vector<u8>): address {
```
hardsetting, line 69: Wouldn't "authentication key" be the correct term here as opposed to "account address"?
In most cases it's the same as the account's address, but if an account is rotated to a new derivable abstract public key this wouldn't work

aips/aip-113.md
```diff
@@ -45,67 +50,66 @@ Derivable AA introduces the following features:
...
+    fun authenticate(
+        account: signer,
+        func_info: FunctionInfo,
+        signing_data: AbstractionAuthData,
+    ): signer acquires DispatchableAuthenticator, DerivableDispatchableAuthenticator {
+        let master_signer_addr = signer::address_of(&account);
+
+        if (signing_data.is_derivable()) {
+            assert!(features::is_derivable_account_abstraction_enabled(), error::invalid_state(EDERIVABLE_ACCOUNT_ABSTRACTION_NOT_ENABLED));
+            assert!(master_signer_addr == derive_account_address(func_info, signing_data.derivable_abstract_public_key()), error::invalid_state(EINCONSISTENT_SIGNER_ADDRESS));
```
hardsetting, line 95: and here we should check for key rotation?
(we can implement later, but just checking if it makes sense as a concept)
```move
let auth_key = derive_auth_key(func_info, signing_data.derivable_abstract_public_key());
let account_address = lookup_original_account_address(auth_key);
assert!(master_signer_addr == account_address, error::invalid_state(EINCONSISTENT_SIGNER_ADDRESS));
```

hardsetting (approved):

alinush (commented):
I think it would be very helpful to show an example of a function that performs DAA signature validation, as an example.

Otherwise, from this AIP, it's not even clear what the declaration of this function is (args, return types, etc.)

aips/aip-113.md
igor-aptos: this is account address - which cannot be changed. so you permanently need to use fixed abstract_public_key.

For rotation:
- We currently don't support "disconnecting" account from original public key, though we can add that.
- You can account::rotate_authentication_key_call, and add another native key to it.
- implementation could support other rotations internally, SIWS implementation doesn't support that

aips/aip-113.md
igor-aptos: hmmm, not sure we want rotation as a concept - wallet would not be able to know/derive the old address. but if we do - we can have that as a separate AIP

sherry-x (approved):

PR #581, opened 2025-04-08
x-chain DAA authentication - Sign-in-with-Solana

igor-aptos:


hardsetting (approved):

aips/xchain_daa_authentication.md
```diff
@@ -0,0 +1,132 @@
+---
+aip: TBD
+title: x-chain DAA authentication using Sign-in-With-Solana
+author: igor-aptos, brian, 0xmaayan, lightmark
+Status: Draft
+type: Framework
+created: 04/08/2024
+---
+
+# AIP-X - (AIP title)
```
hardsetting, line 10: This needs a proper title and number

aips/xchain_daa_authentication.md
```diff
@@ -0,0 +1,132 @@
+---
+aip: TBD
+title: x-chain DAA authentication using Sign-in-With-Solana
+author: igor-aptos, brian, 0xmaayan, lightmark
```
hardsetting, line 4: Any chance I could get added to this? :D

alinush (commented):
I think it would be very helpful to actually show the Move function that performs the DAA signature validation, checking the SIWS signature and matching the `domain` in the SIWS message with the `domain` in the DAA abstract PK.

igor-aptos:
@hardsetting @alinush  - updated AIP based on your comments.

also made appropriate rename in #16345

alinush (commented):
> "If Sign-in-with-Solana is used in the Wallet, domain is verified, and warning is given to the user. If user ignores the warning - they will allow diffent (potentially malicious) websites to execute transactions"

The first sentence is not a sentence. Can you please re-write and state the risk clearly?

alinush (commented):
> If sign_message with above payload is used directly (which is generally discouraged from users to sign), domain will not be verified, and it is on the user to understand risks

This sounds very scary. I thought we would have no such risks.

aips/xchain_daa_authentication.md
igor-aptos: next one is currently aip-121 - https://github.com/aptos-foundation/AIPs/tree/main/aips

you need to rename the file as well, and create an issue and put a link to it in the template

PR #589, opened 2025-04-24
AIP: Validator Priority Fees

guy-goren:
This AIP proposes a mechanism for Validator Priority Fees in Aptos.
The motivation, implementation outline, and rationale are detailed in the markdown file under aips/Validator Priority Fees.md

### DIFF
--- a/aips/aip-126.md
+++ b/aips/aip-126.md
@@ -1,5 +1,5 @@
 ---
-aip: (this is determined by the AIP Manager, leave it empty when drafting)
+aip: 126
 title: Validator Priority Fees
 author: guy-goren, Guoteng Rao, Zekun Li
 Status: Draft
@@ -9,12 +9,8 @@ updated (*optional): <mm/dd/yyyy>
 requires (*optional): <AIP number(s)>
 ---
 
-# AIP-X - Validator Priority Fees
+# AIP-126 - Validator Priority Fees
   
-(Please give a temporary file name to your AIP when first drafting it, such as `aip-x.md`. The AIP manager will assign a number to it after reviewing.)
-
-(Please remove the questions in the "quote box". Provide complete context to the questions being asked in the content that you provide.)
-
 ## Summary
 
 There is currently no mechanism in the Aptos protocol that incentivizes validators to prioritize higher-priced transactions. As a result, no built-in priority fee market exists, which encourages the emergence of informal or "black market" mechanisms for transaction ordering. This becomes increasingly problematic as trading use cases grow, particularly those involving arbitrage opportunities, where fair and open competition over ordering is essential.


MoonShiesty:
i want to start my review by saying i think the AIP moves things in the right direction, and i would support this AIP over the status quo

however i disagree with the motivation and goal of this AIP, specifically:

> side channels benefit participants with privileged access and can erode fairness by creating unequal conditions for different users

the premise that side-channels benefit privileged participants is not sound. i want a priority fee split because it would allow validators to build value generating side channels

the most successful example of side-channels are "MEV-protect" [[1]](https://docs.flashbots.net/flashbots-protect/overview) [[2]](https://jito-foundation.gitbook.io/mev/solana-mev/jitos-mev-priorities) products which bundle user transactions and then auction those bundles to searchers who can execute backruns against the bundled transactions

this approach has been show to benefit all ecosystem participants at the expense of "privileged participants" (ie searchers), because:

 - users transactions are protected from being exposed to frontrunning while lingering in a mempool
 - validators earn tips from bundles, which means validators extract REV instead of searchers. this benefits stakers and validators at the expense of searchers 
 - applications can internalize the value from their orderflow
 - failed transactions don't land on-chain
 
image

rather than framing this proposal as a mechanism to disincentivize side-channels, i would frame this proposal as an incentive against censorship resistance

as currently proposed, validators have zero-incentive not to censor min fee transactions. if the leader wants to achieve a specific order or land their transaction first, the economics for this AIP are for the leader to censor any non-validator transactions from the block

i would prefer to see both alternative solutions (2) and (3) implemented:

a 50/50 fee split provides an economic reason for the leader include other, "less-privileged" participants transactions

and i believe an external block-building API is the ideal solution for building efficient blocks that maximizes staker REV. but this is out-of-scope of this AIP. a 50/50 fee split is a good starting point, and external-block building APIs can always be shipped in a custom client without any consensus changes or need for a vote

MoonShiesty:
i want to start my review by saying i think the AIP moves things in the right direction, and i would support this AIP over the status quo

however i disagree with the motivation and goal of this AIP, specifically:

> side channels benefit participants with privileged access and can erode fairness by creating unequal conditions for different users

the premise that side-channels benefit privileged participants is not sound. i want a priority fee split because it would allow validators to build value generating side channels

the most successful example of side-channels are "MEV-protect" [[1]](https://docs.flashbots.net/flashbots-protect/overview) [[2]](https://jito-foundation.gitbook.io/mev/solana-mev/jitos-mev-priorities) products which bundle user transactions and then auction those bundles to searchers who can execute backruns against the bundled transactions

this approach has been show to benefit all ecosystem participants at the expense of "privileged participants" (ie searchers), because:

 - users transactions are protected from being exposed to frontrunning while lingering in a mempool
 - validators earn tips from bundles, which means validators extract REV instead of searchers. this benefits stakers and validators at the expense of searchers 
 - applications can internalize the value from their orderflow
 - failed transactions don't land on-chain
 
image

rather than framing this proposal as a mechanism to disincentivize side-channels, i would frame this proposal as an incentive against censorship resistance

as currently proposed, validators have zero-incentive not to censor min fee transactions. if the leader wants to achieve a specific order or land their transaction first, the economics for this AIP are for the leader to censor any non-validator transactions from the block

i would prefer to see both alternative solutions (2) and (3) implemented:

a 50/50 fee split provides an economic reason for the leader include other, "less-privileged" participants transactions

and i believe an external block-building API is the ideal solution for building efficient blocks that maximizes staker REV. but this is out-of-scope of this AIP. a 50/50 fee split is a good starting point, and external-block building APIs can always be shipped in a custom client without any consensus changes or need for a vote

MoonShiesty (commented):
i want to start my review by saying i think the AIP moves things in the right direction, and i would support this AIP over the status quo

however i disagree with the motivation and goal of this AIP, specifically:

> side channels benefit participants with privileged access and can erode fairness by creating unequal conditions for different users

the premise that side-channels benefit privileged participants is not sound. i want a priority fee split because it would allow validators to build value generating side channels

the most successful example of side-channels are "MEV-protect" [[1]](https://docs.flashbots.net/flashbots-protect/overview) [[2]](https://jito-foundation.gitbook.io/mev/solana-mev/jitos-mev-priorities) products which bundle user transactions and then auction those bundles to searchers who can execute backruns against the bundled transactions

this approach has been show to benefit all ecosystem participants at the expense of "privileged participants" (ie searchers), because:

 - users transactions are protected from being exposed to frontrunning while lingering in a mempool
 - validators earn tips from bundles, which means validators extract REV instead of searchers. this benefits stakers and validators at the expense of searchers 
 - applications can internalize the value from their orderflow
 - failed transactions don't land on-chain
 
image

rather than framing this proposal as a mechanism to disincentivize side-channels, i would frame this proposal as an incentive against censorship resistance

as currently proposed, validators have zero-incentive not to censor min fee transactions. if the leader wants to achieve a specific order or land their transaction first, the economics for this AIP are for the leader to censor any non-validator transactions from the block

i would prefer to see both alternative solutions (2) and (3) implemented:

a 50/50 fee split provides an economic reason for the leader include other, "less-privileged" participants transactions

and i believe an external block-building API is the ideal solution for building efficient blocks that maximizes staker REV. but this is out-of-scope of this AIP. a 50/50 fee split is a good starting point, and external-block building APIs can always be shipped in a custom client without any consensus changes or need for a vote

PR #610, opened 2025-05-29
Weighted Reward Distribution Based on Activity and Lock Duration

Gta103:
This AIP proposes a weighted reward distribution system that allocates rewards to users based on their activity level and the length of time their assets are locked. By combining these two factors, the mechanism incentivizes both active participation and long-term commitment, resulting in a fairer and more effective rewards allocation. This approach aims to boost network engagement and stability by rewarding users who contribute consistently and stake their assets for extended periods.

### DIFF
--- a/Weighted Reward Distribution Based on Activity and Lock Duration.md
+++ b/Weighted Reward Distribution Based on Activity and Lock Duration.md
@@ -1,0 +1,100 @@
+# AIP: Weighted Reward Distribution Based on Activity and Lock Duration
+
+**Author**: Ali ([@BoulkemhA](https://x.com/BoulkemhA))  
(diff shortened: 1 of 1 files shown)


MoonShiesty:
i like the direction of this AIP

i know we could implement a way to measure staking duration but im not sure how feasible it would be to measure active participation

> active participation

how do we measure active participation?

another approach could be to rewards protocols with gas rebates for transactions they originate

this would give apps with large transaction flow like aggregators an incentive to originate more transactions

Gta103:
Yes I add this file As another api [](https://github.com/Gta103/aptos-aips-from-me/commit/2ba8ff11b4951a497bc3c4b9d526e18327410764)

PR #636, opened 2025-10-22
add contents to aips/vm-performance.md

rahxephon89:


aips/vm-performance.md
```diff
@@ -98,7 +98,51 @@ Due to time constraints, engineering effort and security considerations, the alt
...
+        }
+    }
+```
+
+However, directly replacing the old implementation with this new one is not feasible for two reasons:
+
+- Backward compatibility: modifying the aptos-framework in an incompatible way is generally not allowed.
+- Ecosystem dependencies: existing indexer services and downstream ecosystem projects rely on the current JSON representation of Option. We must continue supporting the legacy format even after introducing the enum version.
+
+To enable a smooth migration toward the new Option representation, we adopt a two-step process:
```
georgemitenkov: , a ... process is adopted:

PR #667, opened 2026-04-09
[proposal] Multisig account timelock

gregnazario:
This proposes an opt-in multisig account timelock.  Similar to Aptos governance, where voting must take place for X time, before the propoosal can be completed.

### DIFF
--- a/aips/aip-145-multisig-account-timelock.md
+++ b/aips/aip-145-multisig-account-timelock.md
@@ -1,0 +1,208 @@
+---
+aip: 145
+title: Multisig Account Timelock
+author: gregnazario (https://github.com/gregnazario)
+discussions-to: <TODO>
+Status: Draft
+type: Standard (Framework)
+created: 04/09/2026
+updated:
+requires: AIP-12, AIP-77
+---
+
+# AIP-145 - Multisig Account Timelock
+
+## Summary
+
+This AIP proposes adding an opt-in timelock mechanism to Aptos multisig accounts. Once a proposed transaction reaches approval quorum, execution is delayed by a configurable period (1 hour to 14 days). A separate, higher signature threshold can be configured to bypass the timelock and allow immediate execution when urgency demands it.
+
+## Motivation
+
+Multisig accounts on Aptos are used to manage treasuries, upgrade smart contracts, and govern DAOs. Today, once a transaction reaches the required number of approvals, any owner can execute it immediately. This creates risk in several scenarios:
+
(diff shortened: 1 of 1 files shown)


aips/aip-145-multisig-account-timelock.md
```diff
@@ -0,0 +1,208 @@
...
+
+```move
+entry fun update_timelock(
+    multisig_account: &signer,
+    new_timelock_secs: u64,
+    new_bypass_num_signatures: u64,
+) acquires MultisigAccount, MultisigAccountTimelock;
+```
+
+**To enable**: `new_timelock_secs > 0` and `new_bypass_num_signatures > num_signatures_required` and `new_bypass_num_signatures <= num_owners`.
```
sherry-x: I guess if a multisig is set to 3/3, with 24 hours of timelock, in this case, they would have no ability to bypass timelock, but should still be allowed to do so, right? Or do we want to block such use case?

aips/aip-145-multisig-account-timelock.md
```diff
@@ -0,0 +1,208 @@
...
+created: 04/09/2026
+updated:
+requires: AIP-12, AIP-77
+---
+
+# AIP-145 - Multisig Account Timelock
+
+## Summary
+
+This AIP proposes adding an opt-in timelock mechanism to Aptos multisig accounts. Once a proposed transaction reaches approval quorum, execution is delayed by a configurable period (1 hour to 14 days). A separate, higher signature threshold can be configured to bypass the timelock and allow immediate execution when urgency demands it.
```
sherry-x: > 1 hour to 14 days

the range sounds good to me generally but not sure if it would be better to allow granularity on the minutes

sherry-x (approved):

aips/aip-145-multisig-account-timelock.md
gregnazario: Granularity is on the second level, this is just to make sure you don't make a timelock for 1 second, or for 30 years

aips/aip-145-multisig-account-timelock.md
gregnazario: In this mechanism, the bypass MUST be more than the normal signatures.  If it's not more than the normal signatures, it defeats the purpose, as the normal number of signatures can always override

aips/aip-145-multisig-account-timelock.md
sherry-x: Gotcha. Let’s do minimal of 1 min and longest 14 days?

PR #668, opened 2026-04-15
[AIP-146] Staking-Based Transaction Limits

georgemitenkov:
AIP for high-limit txns.

### DIFF
--- a/aips/aip-146-staking-based-transaction-limits.md
+++ b/aips/aip-146-staking-based-transaction-limits.md
@@ -7,11 +7,20 @@ Status: Draft
 last-call-end-date (*optional):
 type: Standard (Core, Framework)
 created: 04/14/2026
-updated (*optional): 04/15/2026
+updated (*optional): 04/19/2026
 requires (*optional):
 ---
 
 # AIP-146 - Staking-Based Transaction Limits
+
+***Experimental Feature Disclaimer:***
+*This feature is being introduced on an experimental basis.*
+*The staking-based transaction limits mechanism may be modified, restricted, or removed at any time if it introduces risks to network performance, security, or fairness.*
+*This includes, but is not limited to 1) adjusting minimum stake requirements, 2) changing multiplier tiers, 3) revoking access to elevated limits, 4) disabling the feature entirely.*
+*As such, developers should not build systems that critically depend on guaranteed access to higher transaction limits.*
+*In the event of abuse or unintended consequences, the network may enforce changes or revoke privileges without prior notice.*
+*Future iterations of this feature may also introduce different mechanisms or requirements for accessing elevated limits.*
+*This feature is intended to be iterated on based on real-world usage and feedback.*
 
 ## Summary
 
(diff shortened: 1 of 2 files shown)


aips/aip-transaction-limits.md
```diff
@@ -0,0 +1,308 @@
...
+# AIP-X - Staking-Based Transaction Limits
+
+## Summary
+
+Today, all Aptos transactions share the same execution gas limits.
+However, some workloads such as complex DEX operations (liquidations), large state migrations, and emergency updates need more compute than the standard limit allows.
+Right now, there is no mechanism to request it.
+
+This AIP introduces an opt-in mechanism for higher limits, backed by staking proof.
+Users who prove they control significant stake can request having higher multipliers (above 1x and up to 100x) on gas schedule limits (which otherwise default to 1x).
```
sherry-x: maybe add a bit of reasoning on why we think this make sense: 

Large APT stakers are inherently aligned with the network’s long-term health. Transaction and gas limits are designed to protect the system from spam and maintain high performance. However, large stakers are more likely to have legitimate needs for higher limits rather than malicious intent. Allowing elevated limits—backed by staking proof—strikes a practical balance: it enables more complex on-chain use cases while preserving network performance and resilience.

sherry-x (approved):

aips/aip-transaction-limits.md
georgemitenkov: Done, thanks!

aips/aip-146-staking-based-transaction-limits.md
```diff
@@ -0,0 +1,314 @@
...
+friend fun initialize(
+    aptos_framework: &signer,
+    execution_tiers: vector<TxnLimitTier>,
+    io_tiers: vector<TxnLimitTier>,
+)
+```
+
+Genesis uses the following default tiers:
+
+| Tier | Multiplier   | Execution min stake | IO min stake |
```
sherry-x, line 167: proposing we start a bit small here (more achievable) and increase the requirement if we see abuse or other performance concerns. Something like:

| Tier | Multiplier   | Execution min stake | IO min stake |
|------|--------------|---------------------|--------------|
| 1    | 2x (200 bps) | 1M APT              | 5M APT       |
| 2    | 4x (400 bps) | 5M APT             | 10M APT      |
| 3    | 8x (800 bps) | 10M APT             | 20M APT     |

aips/aip-146-staking-based-transaction-limits.md
```diff
@@ -0,0 +1,314 @@
...
+For the staking proof, the following options are supported: **stake pool ownership**, **delegated voter status**, and **delegation pool delegation**.
+Eligibility is determined by matching the sender's committed stake against a governance-configurable vector of tiers.
+All staking-backed requests also require a **10x minimum gas unit price**.
+
+### Out of scope
+
+- SDK and CLI support for constructing high-limit transactions are not included in this change and are tracked separately.
+- High-limit transactions are not treated differently by mempool.
+
+## High-level Overview
```
sherry-x, line 48: add a disclaimer since we may tweak this feature based on how it's being used initially.

## ⚠️ Experimental Feature Disclaimer

This feature is being introduced on an **experimental basis**.

The staking-based transaction limits mechanism may be **modified, restricted, or removed at any time** if it introduces risks to network performance, security, or fairness. This includes, but is not limited to:
- Adjusting minimum stake requirements
- Changing multiplier tiers
- Revoking access to elevated limits
- Disabling the feature entirely

As such, developers **should not build systems that critically depend on guaranteed access** to higher transaction limits.

In the event of abuse or unintended consequences, the network may enforce changes or revoke privileges without prior notice. Future iterations of this feature may also introduce different mechanisms or requirements for accessing elevated limits.

This feature is intended to be iterated on based on real-world usage and feedback.

georgemitenkov: Added

